PR #14, opened 2016-02-11
Cache

uniquoooo:
- Implemented Caching into DiscordPHP.
    - At the moment, only caches HTTP requests.


uniquoooo:
potato, i want to merge into develop

PR #34, opened 2016-03-18
Make Guzzle Helper options configurable

GIT-RickVH:
Make the Guzzle helper options configurable. Needed this for turning off "verify" for local testing.

Small issue I know, but it is annoying me :P


uniquoooo:
Awesome, thanks!

PR #43, opened 2016-03-26
Adding option resolver, and updating examples to use tokens

aequasi:


uniquoooo:
Looks awesome, thank you!

Just a couple things:
- Would you be able to add backwards-compatibility so that people can just pass through the token? Possibly just remove the `array` type-hinting and have: `(! is_array($options)) ? $options = ['token' => $options];`
- We are also dropping the caching of tokens since the tokens only change if you make it change and we can't get it from an endpoint (not sure if you are using this for your caching pull request or not).

PR #53, opened 2016-03-29
	Updating events to use promises

aequasi:


src/Discord/Parts/Channel/Message.php
```diff
@@ -90,10 +91,15 @@ public function getChannelAttribute()
         );
     }
 
+    public function getFullChannel(Callable $callback)
```
uniquoooo: Any reason this is a callback instead of a promise?

PR #98, opened 2016-04-15
fixed unable to send private message

ivanpepelko:
private message channels don't have type attribute for some reason, Channel::getType() returns 'text' in that case


aequasi:
Can you re-send this to `develop`?

ivanpepelko:
sent to develop

PR #102, opened 2016-04-19
Allow >=4.2 versions of illuminate/support

Sleavely:
I've sifted through DiscordPHPs files where Illuminate\Support is used and there shouldn't be any reason to run `^5.1`.

As far as I can tell there is no breakage, but I'm expecting Travis to have my back in case I'm wrong.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     "require":     {
         "cache/cache":              "^0.1",
         "guzzlehttp/guzzle":        "~5.3|~6.0",
-        "illuminate/support":       "^5.1",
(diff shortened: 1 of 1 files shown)


Sleavely:
No Travis on pull requests? :(

PS. StyleCI reports failure on `src/Discord/Parts/Channel/Channel.php` which is untouched by this PR.

aequasi:
I'm gonna update the other deps to be more like the following, Would you mind updating yours too:

`^4.0|^5.0`

aequasi:
And yeah, checks are just all broke right now

aequasi:
ALSO, close and PR in to develop

PR #113, opened 2016-05-18
Develop

uniquoooo:
- Converted ByteBuffer to v0.3.
- Implemented the `getAll(query)` function on the cache.
- Implemented global users and channel caches through `$discord->channels` and `$discord->users`.
- Implemented sharding of the client.
- Implemented bulk deletion of messages.
- Implemented the `Channel->getMessageHistory()` function.
- Added the `mention_roles` attribute to `Message`.
- Converted the guild asset links to use CDN.
- Converted `manage_roles` to `administrator` and added `change_nickname`, `manage_nicknames` and the new `manage_roles` to permissions.
- Added the `Game` part.
- Fixed up some issues with caching.
- Fixed an issue with ETF.
- Added the `WebSocket->updatePresence()` function.
- Updated DCA to a custom version and added 32-bit OS support.
- Removed caching from invites.
- Disabled ETF by default.
- Cleaned up sending files.


composer.json
```diff
@@ -9,15 +9,16 @@
         }
     ],
     "require":     {
-        "cache/cache":              "^0.1",
```
aequasi: 👍

aequasi:
👍

PR #127, opened 2016-07-04
Avoid double define for DISCORD_TOKEN

SiR-DanieL:
If `DISCORD_TOKEN` is already defied, there's no need to define it again.
It will throw a PHP error.


aequasi:
@uniquoooo is this already done in v4?

uniquoooo:
We don’t use constants in v4, so yeah.

PR #147, opened 2016-08-22
Fix type for user attribute on member

danielholmes:


uniquoooo:
Could you resend this PR to the `develop` branch please? Thanks!

PR #177, opened 2016-11-12
Hide the token while debugging

Xety:
This is done to protect the token credentials, which could be accidentally shown in the console.

### DIFF
--- a/src/Discord/Discord.php
+++ b/src/Discord/Discord.php
@@ -1353,6 +1353,8 @@ class Discord
         $replace = array_intersect_key($secrets, $this->options);
         $config = $replace + $this->options;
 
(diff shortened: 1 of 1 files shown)


uniquoooo:
Could you remove the `loop`, `cachePool` and `logger` attributes from the config? An `unset` statement would be fine.

Xety:
Sure.
Now the result look like this :
```php
object(Discord\Discord) {

        'token' => '*****',
        'bot' => true,
        'loggerLevel' => (int) 200,
        'logging' => true,
        'loadAllMembers' => false,
        'disabledEvents' => [],
        'pmChannels' => true,
        'storeMessages' => false,
        'retrieveBans' => true

}
```

uniquoooo:
Thanks!

PR #209, opened 2017-02-13
Updated requirements to fix Timestamp error

ScrummerLP:
The following error should be fixed now with the update of the requirements:

```PHP Warning:  DateTime::createFromFormat(): It is not safe to rely on the system's timezone settings. You are *required* to use the date.timezone setting or the date_default_timezone_set() function. In case you used any of those methods and you are still getting this warning, you most likely misspelled the timezone identifier. We selected the timezone 'UTC' for now, but please set date.timezone to select your timezone. in /Users/jpache/DEV/web/php-discordbot/src/lib/cache/adapter-common/CacheItem.php on line 157```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
   ],
   "require": {
     "php": "^5.6|^7.0",
-    "cache/array-adapter": "^0.4.0",
+    "cache/array-adapter": "^0.4.2",
     "guzzlehttp/guzzle": "~5.3|~6.0",
     "illuminate/support": "^4.0|^5.0",
     "nesbot/carbon": "^1.18",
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: composer.json [hunk 1]
```diff
@@ -14,7 +14,7 @@
   ],
   "require": {
     "php": "^5.6|^7.0",
-    "cache/array-adapter": "^0.4.0",
+    "cache/array-adapter": "^0.4.2",
```
ScrummerLP: Should be updated! As you can see here: https://github.com/php-cache/adapter-common/commit/a1361f7f8b8557f6e6392131f8078a189e345b40#diff-fbe933f438ac3d91fb49186a962c4978

The Caching has been switched from DateTime to Timestamp. This should be updated to prevent errors

uniquoooo:
Could you resend this PR to the `develop` branch please?

PR #211, opened 2017-02-14
Search Ability (USER ONLY) for Guild Part

donkkill13:
I used StyleCI this time, I think I did good? :D


donkkill13:
Only users can use this endpoint, I just figured out, so it's still useful if you use a UserToken and put `'bot' => false` into the discord options

I completely forgot to check if bots can use this endpoint before making this request lol

aequasi:
Should rename this PR

Issue #174, opened 2016-11-08
Memory leak


Xety:
Hello,

I was doing my bot when i did for fun a cmd to display the memory used by the script and... i noticed a huge memory leak.
I thought it was my script, but i tried with the exemple code and the memory leak is still there :

Code used to test it :
``` php
<?php
//Set the memory limit to 6MB else it will take forever to test it.
ini_set('memory_limit', '6M');

include __DIR__.'/vendor/autoload.php';

$discord = new \Discord\Discord([
    'token' => 'your-auth-token',
]);

$discord->on('ready', function ($discord) {
    echo "Bot is ready.", PHP_EOL;

    // Listen for events here
    $discord->on('message', function ($message) {

        $memoryKo = round(memory_get_usage() / 1024);
        $memoryMo = number_format($memoryKo / 1024, 1, ',', ' ');

        $message->channel->sendMessage('Memory used : `' . $memoryKo . 'Ko` (`' . $memoryMo . 'Mo`).');


        echo "Recieved a message from {$message->author->username}: {$message->content}", PHP_EOL;
    });
});

$discord->run();
```

Write a message in a channel and you will see the memory leak. 

![887ae1350b3dfea1b218f64ae2e0601e1](https://cloud.githubusercontent.com/assets/8210023/20115999/ce5ef838-a5fa-11e6-9b1c-8cdc3aeadb07.png)

I don't know if this memory leak come from this library or from another one.

sciencefyll:
I've noticed this as well. We made a "hack fix" by restarting the bot every now and then, otherwise everything would crash 0.o

sciencefyll:
Might be: DiscordPHP/src/Discord/Discord.php $logger.

@aequasi @Sleavely @xKairu @danielholmes

Sleavely:
Is the bot connected to voice chat? In that case it might be what @sciencefyll mentioned, the [voice logger being initiated](https://github.com/teamreflex/DiscordPHP/blob/b44c4be11281827ec493c80d1a319eca3f46d849/src/Discord/Discord.php#L1060) over and over.

[...]

PR #212, opened 2017-02-18
Updated a lot of ->pull() functions to ->offsetSet().

donkkill13:
Also fixed some Repositories that had $class instead of $part, thus failing when using ->fetch($id) on them.

### DIFF
--- a/src/Discord/Parts/Channel/Message.php
+++ b/src/Discord/Parts/Channel/Message.php
@@ -71,7 +71,8 @@ class Message extends Part
         'nonce',
         'mention_roles',
         'pinned',
-        'reactions',
+		'reactions',
+		'webhook_id',
     ];
 
     /**
@@ -83,7 +84,8 @@ class Message extends Part
      */
     public function reply($text)
     {
-        return $this->channel->sendMessage("{$this->author}, {$text}");
+		$user = $this->getAuthorAttribute(1);
+        return $this->channel->sendMessage("{$user->id}, {$text}");
     }
 
     /**
@@ -180,13 +182,13 @@ class Message extends Part
     {
         $roles = new Collection([], 'id');
 
-        $guildRoles = $this->channel->guild->roles;
-
-        foreach ($this->attributes['mention_roles'] as $mentionRoleID) {
-            if ($guildRoles->has($mentionRoleID)) {
-                $roles->push($guildRoles->offsetGet($mentionRoleID));
-            }
-        }
+		$guildRoles = $this->channel->guild->roles;
+		
+		foreach ($this->attributes['mention_roles'] as $mentionRoleID) {
+			if ($guildRoles->has($mentionRoleID)) {
+				$roles->push($guildRoles->offsetGet($mentionRoleID));
+			}
+		}
 
         return $roles;
     }
@@ -214,17 +216,16 @@ class Message extends Part
      */
     public function getAuthorAttribute($type = 0)
     {
-        if ($this->channel->type != Channel::TYPE_TEXT || $type === 1 || ! $this->channel->guild->members->has($this->attributes['author']->id)) {
+        if ($this->channel->type != Channel::TYPE_TEXT || $type === 1 || !$this->channel->guild->members->has($this->attributes['author']->id)) {
             if ($this->attributes['author'] instanceof User) {
-                return $this->attributes['author'];
-            } else {
-                $this->attributes['author'] = $this->factory->create(User::class, $this->attributes['author'], true);
-
-                return $this->attributes['author'];
-            }
+				return $this->attributes['author'];
+			} else {
+				$this->attributes['author'] = $this->factory->create(User::class, $this->attributes['author'], true);
+				return $this->attributes['author'];
+			}
         }
 
-        return $this->channel->guild->members->offsetGet($this->attributes['author']->id);
+		return $this->channel->guild->members->offsetGet($this->attributes['author']->id);
     }
 
     /**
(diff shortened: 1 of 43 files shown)


donkkill13:
I also added checks on message update, sendMessage, etc.. to see if the user has 'storeMessages' as true before storing. Before it would just store those messages, so every Message Update would store a message, resulting in unwanted ram usage. (It would store duplicates as well)

Basically, this entire update removes duplicates in EVERYTHING.

$discord->users went from 130+ to just 41 after applying this update, 30 of the 130+ was just my account!

https://nig.tools/3dyns508.txt (130+) to https://nig.tools/hi1zp80f.txt (41)

Also applyed a fix on `->isSpeaking()` in the `Voice.php`, added $embed support to `$channel->sendFile()`, added a check to see if user is a bot before making an application, ability to update a message's embed (add an embed or edit an embed), removed `->sendFile()` from `guzzle.php`, changed some repositories so `->fetch()` actually works on them (changed $class to $part on their file, and added, in `->fetch()`, if `['guild_id']` exists, add it in (since `->fetch()` on a members repository doesn't return guild_id when you require it to make a member)), and probably some more stuff.

(also $state->user_id, not $state->id in VoiceStateUpdate)

OVERALL IMPROVEMENTS, I RECOMMEND!

donkkill13:
I added more fixes, replaced `->get('id', $id)` with `->offsetGet($id)` so it uses less processing power? idk lol

src/Discord/functions.php
```diff
@@ -38,3 +38,18 @@ function mentioned($part, Message $message)
...
 }
+
+/**
+ * Calculates UNIX timestamp from snowflake.
+ *
+ * @param INT $snowflake The snowflake
+ *
+ * @return UNIX timestamp
+ */
+function timestampFromSnowFlake(int $snowflake)
```
uniquoooo: Snowflakes should be in `string` format as they are 64-bit integers.

src/Discord/Parts/Channel/Message.php
```diff
@@ -210,13 +212,19 @@ public function getMentionsAttribute()
      *
      * @return Member|User The member that sent the message. Will return a User object if it is a PM.
      */
-    public function getAuthorAttribute()
+    public function getAuthorAttribute($type = 0)
```
uniquoooo: I'm thinking of possibly splitting the `author` attribute into two attributes, one for the User part and one for the Member part but I will come back to it later.

donkkill13: I just did a quick thing because if you have all users/members loaded, and you're in a guild with 1k+ members,  then you do $message->author, it'll search through 1k+ members to get the author attribute.


I think splitting it would be the best course of action though, have 'author' and 'member', or just have 'author' spit out a user and have people go through `$message->channel->guild->members->offsetGet($message->author->id)` for the `member` attribute

src/Discord/Parts/User/User.php
```diff
@@ -72,7 +71,7 @@ public function sendMessage($message, $tts = false, $embed = null)
         $deferred = new Deferred();
 
         $this->getPrivateChannel()->then(function ($channel) use ($message, $tts, $embed, $deferred) {
-            $channel->sendMessage($message, $tts, $embed)->then(function ($response) use ($deferred) {
+            $channel->sendMessage($message, $tts, $embed)->then(function ($response) use ($channel, $deferred) {
```
uniquoooo: Any reason for the `use ($channel)`?

donkkill13: Nope, it's left over from when I was storing the message after creating it (I didn't know at the time about a MESSAGE_CREATE event being fired after creating the message lol

donkkill13:
Updated it more (I also added the ability of the guild filling it's emoji repository, which was not the case before)

src/Discord/functions.php
```diff
@@ -38,3 +38,15 @@ function mentioned($part, Message $message)
...
+/**
+ * Calculates UNIX timestamp from snowflake.
+ *
+ * @param STRING $snowflake The snowflake
+ *
+ * @return UNIX timestamp
+ */
+function timestampFromSnowFlake(string $snowflake)
+{
+    return (($snowflake >> 22) + 1420070400000);
```
uniquoooo: This constant will also need to be a string as it is more than the PHP max integer size on 32-bit PHP so we need to find a way to calculate timestamps without using it.

donkkill13: It returns a timestamp with microseconds, so to resolve the UNIXtime, you get the Timestamp w/o MicroSeconds you'd do floor($timestamp / 1000). I have yet to find a way to resolve the Timestamp w/ MicroSeconds using Carbon though.. (I used http://i.imgur.com/UxWvdYD.png as a guide)

src/Discord/functions.php
```diff
@@ -38,3 +38,17 @@ function mentioned($part, Message $message)
         return strpos($message->content, $part) !== false;
     }
 }
+
+/**
+ * Calculates UNIX timestamp from snowflake.
+ *
+ * @param STRING $snowflake The snowflake
```
aequasi: return type should be `string`

src/Discord/functions.php
```diff
@@ -38,3 +38,17 @@ function mentioned($part, Message $message)
...
+ *
+ * @return UNIX timestamp
```
aequasi: return type should be `float|int`

src/Discord/functions.php
```diff
@@ -38,3 +38,17 @@ function mentioned($part, Message $message)
...
+ */
+function timestampFromSnowFlake(string $snowflake, $float = false)
```
aequasi: also, remove scalar type hint. Lib supports php 5.x

Xety:
This PR actually reduce the memory leak (cf #174) from 60% than before. That's a very good new, keep it up. :+1:

PR #299, opened 2020-08-23
Add cooldown option to commands

oliverschloebe:
Helping people calm the f**k down while using bot commands. 🤗

#### Example usage:

```php
$discord->registerCommand('ping', function($message, $args) use ($discord) {
	$message->reply("pong");
}, [
	'cooldown' => 10000,
	'cooldownMessage' => 'T Minus %d' // %d = remaining seconds
]);
```

props @Thorrdu + me adding ability to customize cooldown message


davidcole1340 (commented):
Possibly look at using `microtime()` to get the time as the `time()` function is only accurate to the second? Could use `microtime()` then change the cooldown to be specified in milliseconds. Could you also resolve the conflicts?

src/Discord/CommandClient/Command.php
```diff
@@ -163,6 +190,17 @@ public function handle(Message $message, array $args)
             array_unshift($args, $subCommand);
         }
 
+        $currentTime = round(microtime(true));
```
davidcole1340, line 193: Use `$currentTime = round(microtime(true) * 1000)` to get the current epoch in milliseconds.

davidcole1340 (changes requested):
Try to keep everything in milliseconds, since `microtime` is accurate to the millisecond.

davidcole1340:
Thank you :)

PR #300, opened 2020-08-25
Replace deprecated xkcd parameter with unique parameter

oliverschloebe:
https://discord.com/developers/docs/resources/channel#create-channel-invite


src/Discord/Parts/Channel/Channel.php
```diff
@@ -287,10 +287,10 @@ public function createInvite($max_age = 3600, $max_uses = 0, $temporary = false,
             [
                 'validate' => null,
```
davidcole1340, line 288: Could you remove this field? It is also no longer used.

davidcole1340 (approved):
Thank you 👍

src/Discord/Parts/Channel/Channel.php
davidcole1340: Don't worry, I will do it, going to change this function a bit.

PR #302, opened 2020-08-26
Mitigate missing members and roles for guild

mvaessen:
Fix for when both roles and members are disabled.
This issue happens when sending a message via a private channel.

Thanks for the great work on the bot!


davidcole1340 (approved):

davidcole1340:
Thank you!

PR #343, opened 2020-09-29
Update code to PHP 7.2 type declaration

Ravisdo:
* add proper type declaration
* add missing class variables
* run php cs
* update composer.json to use at least php 7.2


Ravisdo:
Okay, there are a few issues within the code declaration. I will fix this

davidcole1340:
Thank you!

Issue #345, opened 2020-10-01
PHP Notice:  Undefined variable: message in vendor/team-reflex/discord-php/src/Discord/WebSockets/Events/MessageDelete.php on line 33


key2peace:
https://github.com/teamreflex/DiscordPHP/blob/e0f9ebace1efedab432ed685bc3323ddb610f762/src/Discord/WebSockets/Events/MessageDelete.php#L32

perhaps make it
```
    public function handle(Deferred &$deferred, $data)
    {
        $message = null;
        if ($guild = $this->discord->guilds->get('id', $data->guild_id)) {
```

PR #348, opened 2020-10-01
Hotfix/undefined variable message in message delete

Ravisdo:
#345 undefined variable message in MessageDelete.php

+ set message default to null


davidcole1340:
Cheers

PR #360, opened 2020-10-12
Update Event::VoiceStateUpdate / Add Channel::editMessage

key2peace:
Event::VoiceStateUpdate:
- use allowVoice()
- get rid of the foreach and just check if it has() the member
- since we are looping through the channels anyways, why not push him here as well?

Channel::editMessage:
- because it's missing

### DIFF
--- a/src/Discord/Parts/Channel/Channel.php
+++ b/src/Discord/Parts/Channel/Channel.php
@@ -703,6 +703,45 @@ class Channel extends Part
     }
 
     /**
+     * Edit a message in the channel.
+     *
+     * @param Message $message  The message to edit.
+     * @param string  $text     The text to of the message.
+     * @param bool    $tts      Whether the message should be sent with text to speech enabled.
+     * @param Embed|array|null $embed An embed to send.
+     *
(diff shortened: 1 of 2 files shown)


src/Discord/Parts/Channel/Channel.php
```diff
@@ -702,6 +702,45 @@ function ($response) use ($deferred) {
...
+     *
+     * @param Message $message  The message to edit.
+     * @param string  $text     The text to of the message.
+     * @param bool    $tts      Whether the message should be sent with text to speech enabled.
+     * @param Embed|array|null $embed An embed to send.
+     *
+     * @return PromiseInterface
+     * @throws \Exception
+     */
+    public function editMessage(Message $message, string $text, bool $tts = false, $embed = null): PromiseInterface
```
davidcole1340, line 716: To edit a message you are supposed to use `$channel->messages->save($message)` but we can leave this since it's easier.

davidcole1340 (approved):

davidcole1340:
Cheers

PR #363, opened 2020-10-13
Small updates found along the way

key2peace:


src/Discord/Parts/Channel/Channel.php
```diff
@@ -164,7 +164,11 @@ protected function getRecipientsAttribute(): Collection
      */
     protected function getGuildAttribute(): Guild
     {
-        return $this->discord->guilds->get('id', $this->guild_id);
+        if (isset($this->attributes['guild_id'])) {
```
davidcole1340, line 167: Can just leave it as is but need to change the typehinting:
```php
protected function getGuildAttribute(): ?Guild
{
    return $this->discord->guilds->get('id', $this->guild_id);
}
```

davidcole1340: Need the typehinting either way as you return null in your version.

davidcole1340 (changes requested):
Cheers, can you just change the `Channel::getGuildAttribute()` function? It can go back to the way it was as `$this->guild_id` will simply be null, and `->get()` will return null if the guild is not found. Just need to update typehinting.

PR #369, opened 2020-10-17
Add missing activity types

key2peace:


src/Discord/Parts/User/Activity.php
```diff
@@ -38,7 +38,9 @@ class Activity extends Part
     const TYPE_PLAYING = 0; // Playing {$this->name}
     const TYPE_STREAMING = 1; // Streaming {$this->details}
     const TYPE_LISTENING = 2; // Listening to {$this->name}
+    const TYPE_WATCHING = 3; // Watching {$this->name}
```
davidcole1340, line 41: Where did you find that watching is `3`? I can't seem to find it in the docs.

key2peace: activity type 3 is not documented indeed, but it is like type 4 (custom) a client only feature,, bots may not use it. It does however seem to work when you use GAME (which will vanish in v8) instead of ACTIVITY.

start reading at https://discord.com/channels/115233111977099271/234582138740146176/766745908624621678

PR #371, opened 2020-10-19
Since we don't have $discord->channels. a little helper

key2peace:


src/Discord/Discord.php
```diff
@@ -1358,6 +1358,25 @@ public function __set(string $name, $value)
         $this->client->{$name} = $value;
     }
 
+    /**
+     *
```
davidcole1340, line 1362: This line can be removed

src/Discord/Discord.php
```diff
@@ -1358,6 +1358,25 @@ public function __set(string $name, $value)
...
+     * Gets a channel.
+     *
+     * @param string|int $channel_id Id of the channel.
+     *
+     * @return Channel
+     */
+    public function channelGet($channel_id): ?Channel
```
davidcole1340, line 1369: Maybe `getChannel` or `findChannel` instead?

src/Discord/Discord.php
```diff
@@ -1358,6 +1358,25 @@ public function __set(string $name, $value)
...
+    {
+        foreach ($this->guilds as $guild) {
+            if ($channel = $guild->channels->get('id', $channel_id)) {
+                return $channel;
+            }
+        }
+
```
davidcole1340, line 1376: Could also loop through `$this->private_channels`?

davidcole1340 (changes requested):

src/Discord/Discord.php
key2peace: the name was inspired by offsetGet

davidcole1340: I think it should be the other way around, the only reason it is called `offsetGet` is because that is the function name in the PHP interface: https://www.php.net/manual/en/class.arrayaccess.php

davidcole1340 (approved):

PR #373, opened 2020-10-22
More descriptive error outputs

valzargaming:
$e passes useful information that currently is not being taken advantage of. With this change
image
would become 
image


davidcole1340:
Cheers

PR #374, opened 2020-10-22
Make embed constructible with getters/setters 

valzargaming:


valzargaming:
This needs to be double-checked for functionality.

src/Discord/Parts/Embed/Embed.php
```diff
@@ -38,7 +38,104 @@ class Embed extends Part
...
      */
     protected $fillable = ['title', 'type', 'description', 'url', 'timestamp', 'color', 'footer', 'image', 'thumbnail', 'video', 'provider', 'author', 'fields'];
-
+    
+    /**
+     * Constructs a new instance.
+     * @param array  $embed
+     * @throws \Throwable
+     */
+    function __construct(array $embed = array()) {
```
davidcole1340, line 47: Constructor needs to extend the `Part` constructor otherwise no factory, http etc.

src/Discord/Parts/Embed/Embed.php
```diff
@@ -38,7 +38,104 @@ class Embed extends Part
...
+     * @internal
+     */
+    function __get($name) {
+        if (\property_exists($this, $name)) {
+            return $this->$name;
+        }
+        
+        switch ($name) {
+            case 'datetime':
+                return (new \DateTime('@'.$this->timestamp));
```
davidcole1340, line 132: Could use `Carbon` instead of `DateTime`? Same syntax afaik

src/Discord/Parts/Embed/Embed.php
```diff
@@ -182,4 +308,214 @@ private function attributeHelper($key, $class)
...
+     */
+    function setAuthor($name, string $iconurl = '', string $url = '') {
+        $name = (string) $name;
+        
+        if (\strlen($name) === 0) {
+            $this->author = null;
+            return $this;
+        }
+        
+        if (\mb_strlen($name) > 256) {
```
davidcole1340, line 328: Any reason for using `mb_strlen` over `strlen`? mbstring will have to be added to requirements if used.

davidcole1340 (changes requested):
Sorry for taking so long to review, the constructor will need to implement the same footprint as the `Part` constructor otherwise no factory, http etc.

src/Discord/Parts/Embed/Embed.php
valzargaming: IIRC it had something to do with encoding, but it strlen may also work in this context.

src/Discord/Parts/Embed/Embed.php
```diff
@@ -38,7 +38,104 @@ class Embed extends Part
...
+            
+            if (!empty($embed['footer'])) {
+                $this->setFooter(
+                    ((string) ($embed['footer']['text'] ?? '')),
+                    ((string) ($embed['footer']['icon_url'] ?? ''))
+                );
+            }
+            
+            if (!empty($embed['image'])) {
+                $this->image = array(
```
Ravisdo, line 79: could you use also the short array syntax in that file?
`$this->image = [...]`

davidcole1340: `php-cs-fixer` will change that when it's run, so it's fine for now.

src/Discord/Parts/Embed/Embed.php
```diff
@@ -38,7 +38,107 @@ class Embed extends Part
...
      */
     protected $fillable = ['title', 'type', 'description', 'url', 'timestamp', 'color', 'footer', 'image', 'thumbnail', 'video', 'provider', 'author', 'fields'];
-
+    
+    /**
+     * Constructs a new instance.
+     * @param array  $embed
+     * @throws \Throwable
+     */
+    function __construct(array $embed = array()) {
```
davidcole1340, line 47: Needs to implement [this constructor](https://github.com/teamreflex/DiscordPHP/blob/4a7a5218c9e13469cdb223f968a0774b2bfd0512/src/Discord/Parts/Part.php#L123) otherwise part won't work

davidcole1340 (changes requested):
Just need to fix the constructor, cheers!!

davidcole1340 (approved):
I changed it around a little bit, had to remove the constructor because it needs to inherit the parent constructor so that factory works, but can use Embed::fill(). Cheers for the PR!

PR #377, opened 2020-10-23
Cleanups, using the repositories and small updates

key2peace:
- check if we not already have an object before creating a new one.
- changed allowVoice to be similar to allowText in case discord decides to expand.
- (un)muteMember updated.
- deleteMessages, enforce 100 message limit by sending it in batches when neccessary.
- other small cleanups.
- made on('ready') handler in __construct selfdestructing, and moved below disabledEvents (in case someone puts 'ready' in there to temp disable his own scripts.

### DIFF
--- a/src/Discord/Parts/Channel/Channel.php
+++ b/src/Discord/Parts/Channel/Channel.php
@@ -150,10 +150,10 @@ class Channel extends Part
 
         if (array_key_exists('recipients', $this->attributes)) {
             foreach ((array) $this->attributes['recipients'] as $recipient) {
-                if (! $user = $this->discord->users->get('id', $receipient->id)) {
+                if (! $user = $this->discord->users->get('id', $recipient->id)) {
                     $user = $this->factory->create(User::class, (array) $recipient, true);
                 }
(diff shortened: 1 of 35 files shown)


src/Discord/Parts/Channel/Channel.php
```diff
@@ -150,7 +150,10 @@ protected function getRecipientsAttribute(): Collection
 
         if (array_key_exists('recipients', $this->attributes)) {
             foreach ((array) $this->attributes['recipients'] as $recipient) {
-                $recipients->push($this->factory->create(User::class, (array) $recipient, true));
+                if (! $user = $this->discord->users->get('id', $receipient->id)) {
+                    $user = $this->factory->create(User::class, (array) $recipient, true);
+                }
+                $receipients->push($user);
```
oliverschloebe, line 156: typo in `$receipients`?

key2peace: fixed in 9efb575

davidcole1340 (approved):
Cheers

PR #378, opened 2020-10-23
little helper to get the int value of a html like color

key2peace:


davidcole1340 (changes requested):
Hmm, I'm not sure if this belongs in the `Discord` class, maybe move it to the `src/Discord/functions.php` file and set `$colorTable` to be a `const`?

src/Discord/Discord.php
```diff
@@ -290,6 +290,51 @@ class Discord
      */
     protected $client;
 
+    /**
+     * The HTML Color Table.
+     *
+     * @array HTML Color Table.
+     */
+    public $colorTable = [
+        'indianred' => 'cd5c5c', 'lightcoral' => 'f08080', 'salmon' => 'fa8072', 'darksalmon' => 'e9967a',
```
kjc75, line 299: Might also consider storing the values as their hex encoded value and save the string->int conversion via ```hexdec()``` on each use.

You'd then have ```'indianred' => 0xcd5c5c, ...```, and in the ```getColor()``` function you'd simply return ```$this->colorTable[$colorName]``` for found keys.

key2peace: will do that now

key2peace: fd320c2 contains @kjc75's sugestion
69649cb contains @davidcole1340's change request

davidcole1340 (approved):

Issue #381, opened 2020-10-25
Errors are never displayed


kenygamer:
DiscordPHP silences all the errors, making it impossible to know what is failing in the program.

Code to reproduce:
```php
$discord = new Discord(...);
$discord->on("message", function($message){
    throw new \Exception("Test exception"); 
});
```

oliverschloebe:
This is a known issue, see https://github.com/teamreflex/DiscordPHP/issues/358

kenygamer:
> This is a known issue, see #358

It was closed. Is there any fix?

key2peace:
Did you update to the recent version and tried?

kenygamer:
> Did you update to the recent version and tried? (See [8410c57](https://github.com/teamreflex/DiscordPHP/commit/8410c57a23b8a7cb230d9730b893106c8f2b4f27))

Yes, I did.

key2peace:
I think you need to look at vendor/evenement/evenement/src/Evenement/EventEmitterTrait.php after install and wrap the call_user_func's in function emit around a try/catch to see if that does what you want.

kenygamer:
> I think you need to look at vendor/evenement/evenement/src/Evenement/EventEmitterTrait.php after install and wrap the call_user_func's in function emit around a try/catch to see if that does what you want.
> 
> Consider this a workaround for testing. I have an idea...

Thanks, but your PR is still not working.

key2peace:
well, I won't mess in external libraries itself, this is as close as I can get to it

kenygamer:
> well, I won't mess in external libraries itself, this is as close as I can get to it

Does anybody know why this happens?

davidcole1340:
It's an 'issue' with ReactPHP promises (issue in quotations because technically it's not really an issue) where any exceptions that are not handled are swallowed. In promises v2.0 any unhandled exceptions will be thrown like normal exceptions, but we will have to wait a while for that.

For now, I've fixed @key2peace's PR so it should emit the exception and log it in the console: 1ad7758c050a733511cb023a4780647b7c255ee1

Could you try update with `composer update`?

kenygamer:
> It's an 'issue' with ReactPHP promises (issue in quotations because technically it's not really an issue) where any exceptions that are not handled are swallowed. In promises v2.0 any unhandled exceptions will be thrown like normal exceptions, but we will have to wait a while for that.
> 
[...]

PR #384, opened 2020-10-25
wrapping all emit's around a try/catch

key2peace:
putting things a little closer, might help #381


davidcole1340 (changes requested):
Cheers, could you put the trait in a seperate file? Maybe in the helpers folder

davidcole1340 (approved):

PR #387, opened 2020-10-26
small debug update

key2peace:
simply show how many got skipped during the loop because they seem to exist already.


davidcole1340 (approved):
Cheers

PR #388, opened 2020-10-27
Add DiscordPHP version info to default help menu

hemberger:
Improve the default bot description by adding the DiscordPHP version
information (which is displayed in the help menu, if the description
is not customized).

It may or may not be desirable to put the version info somewhere in
the help menu that _cannot_ be customized.


davidcole1340 (approved):
Cheers

PR #389, opened 2020-10-27
changed chunking to send one guild at a time

key2peace:
based on api docs, if using intents, you may only pass 1 guild to get the guild members. you might not use the 'intents' value in the client itself, but you DO probably have intents enabled on the panel if using v6.

noticed a bunch of attempts to get more chunks with empty results, applying this helped.


davidcole1340 (approved):
Cheers

PR #391, opened 2020-10-28
lazy guilds

key2peace:
will work on the caller later, goodnight


key2peace:
after discussion on discord-api about this, I heard it seems to be client only for now,, so I don't need to code the requester yet, feel free to still push it as preparation should the event be made available

Issue #393, opened 2020-10-30
React promises


davidcole1340:
https://github.com/reactphp/promise#done-vs-then

PR #394, opened 2020-10-30
some then to done (#393,#381) and more

key2peace:
- please make sure in your scripts you use done when DONE with a promise instead of then for proper debugging/releasing,
 - remember: ...->done(function ($response) { ok }, function ($response) { fail });
 - chunks: push into $discord->guilds after every chunk, not when all chunks have been received
 - replaced get/push by offset* where possible, david says its faster


key2peace:
will do it over someday

PR #401, opened 2020-11-06
return null if the part does not exist (just in case)

key2peace:
still faster then get, but more sane and less noisy


davidcole1340:
`return $this->items[$offset] ?? null;` ?

key2peace:
done now, however, I didn't do it that way before to keep the code style the same.

davidcole1340 (approved):
Cheers

PR #402, opened 2020-11-06
Update/movetooffset

key2peace:


key2peace:
code has been running stable for more then 5 hours, still running

key2peace:
will do in parts instead

PR #405, opened 2020-11-07
getGuildAttribute can be null

valzargaming:
If the bot is removed from a guild then getGuildAttribute will return null on presence update


davidcole1340 (approved):
Cheers

Issue #408, opened 2020-11-07
Missing channel->bulkDelete() function


valzargaming:
The $channel part should have a ->bulkDelete() function that takes an integer number and tries to fetch and deletes the last number of messages in a channel.

There is currently no equivalent function that currently exists with the MessageRepository or Channel part. We currently must either already have the $message object, know the message ID, or have a ->messages collection to pass. The best usage that currently exists is the channel part ->getMessageHistory() which accepts an array of options with limit => integer number of messages to fetch, then passing that collection object back to the repository to delete them.

valzargaming:
This should work, but could do with review and consideration for a return.

davidcole1340:
All done #410

PR #410, opened 2020-11-07
Add bulkDelete function to Channel part

valzargaming:
Addresses #408


valzargaming:
Tested and confirmed working

davidcole1340 (approved):
Changed the function name since `bulkDelete` is quite broad, so now it is `limitDelete`. Cheers!

PR #418, opened 2020-11-14
Guard against null beingconverted to an empty string

tomalbers:
this causes build in help to fail for example.


davidcole1340 (approved):
Cheers

PR #447, opened 2020-11-29
add allowed_mentions support to Channel::sendMessage

key2peace:
see https://discord.com/developers/docs/resources/channel#allowed-mentions-object on how to utilize


davidcole1340:
Does it still work if `allowed_mentions` is null?

key2peace:
since $message->reply() is doing a sendMessage, and I still get replies from the bot that get highlighted by discord, yes

davidcole1340 (approved):
Cheers

PR #463, opened 2020-12-22
Update/403prevention channels

key2peace:
note, I used NoPermissionsException in order not to break existing scripts that rely on this exception in try/catch statements


davidcole1340 (approved):
Cheers

PR #473, opened 2020-12-30
Update Member's property return types

JaxkDev:
Updates the return types of Member's `premium_since` and `joined_at` property


davidcole1340 (approved):
Cheers

PR #485, opened 2021-01-18
Fix link to class reference

sudofox:
```diff
- http://teamreflex.github.io/DiscordPHP/reference/
+ https://discord-php.github.io/DiscordPHP/reference/
```

old link 404s.

unsure if page-data/index/page-data.json needs to be updated, haven't done much with GitHub pages before.


davidcole1340:
Seems I forgot to rebuild the documentation - I've queued a rebuild so it should be fixed soon. Thank you for reporting!

PR #502, opened 2021-02-25
Implemented the ability to use proper discord replies

svengjohnson:


davidcole1340:
Thanks for the PR! I've already implemented this on the `develop` branch along with the new gateway version 8 features. We aren't adding any new features to the current `master`. If you'd like to use `develop` that instead, you can run `composer require team-reflex/discord-php dev-develop`. Otherwise, we will be releasing it within a week.

PR #520, opened 2021-04-21
Minor updates to docs

hemberger:
* The `docs` workflow only runs on releases, not off the master branch.

* The `loggerLevel` and `httpLogger` options have been removed.

* The `logger` option defaults to log level DEBUG.

### DIFF
--- a/docs/src/pages/api/00_intro.md
+++ b/docs/src/pages/api/00_intro.md
@@ -2,7 +2,7 @@
 title: "Intro"
 ---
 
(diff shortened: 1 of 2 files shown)


hemberger:
For the record, I'm disappointed that the `loggerLevel` option was removed. I felt that was very useful option with a low maintenance cost. Nevertheless, I am very grateful for the v6 update!

key2peace:
from my script:
```php
$logger = new Logger('BOT');
$logger->pushHandler(new StreamHandler(__DIR__ . '/logs/stfc.log', Logger::DEBUG));
```
I defined loglevel in the pushhandler where I also state to put it into a file instead of stdout

davidcole1340 (approved):
Thank you!

PR #521, opened 2021-04-21
Fix use of logging options that were removed in v6

### LINKED ISSUE #519: joinVoiceChannel refers to removed options
**Environment**
- DiscordPHP Version:
    - 6.0.0

**Describe the bug**
The `joinVoiceChannel` function still uses options that have been removed in v6.0.0.
https://github.com/discord-php/DiscordPHP/blob/8e91539ea81367d53f491209a2546d0587483e04/src/Discord/Discord.php#L1167-L1172

I would submit a PR removing them, but I'm not exactly sure how you want to fix this. Use `$this->logger` if the passed in logger is `null` instead? Create a new logger with default settings just for the voice client?

Thanks!

hemberger:
The `logging` and `loggerLevel` options were removed in v6.0.0 by
commit d0fbd3b926. Therefore, if the `$logger` passed to the method
`Discord::joinVoiceChannel` is null (the default), we will now use
the same logger as the `Discord` client.

Fixes #519.

### DIFF
--- a/src/Discord/Discord.php
+++ b/src/Discord/Discord.php
@@ -1118,7 +1118,7 @@ class Discord
      * @param Channel              $channel The channel to join.
      * @param bool                 $mute    Whether you should be mute when you join the channel.
      * @param bool                 $deaf    Whether you should be deaf when you join the channel.
-     * @param LoggerInterface|null $logger  Voice client logger.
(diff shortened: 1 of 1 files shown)


davidcole1340 (approved):
Thank you!

PR #526, opened 2021-04-23
Message link

valzargaming:
Calling $message->link() will provide the message link.

### DIFF
--- a/src/Discord/Parts/Channel/Message.php
+++ b/src/Discord/Parts/Channel/Message.php
@@ -672,10 +672,14 @@ class Message extends Part
     }
     
     /**
-     * {@inheritdoc}
+     * Gets the message link attribute.
+     *
+     * @return String|null
      */
-    public function link(): string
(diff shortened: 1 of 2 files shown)


src/Discord/Parts/Channel/Message.php
```diff
@@ -670,4 +670,12 @@ public function getRepositoryAttributes(): array
             'channel_id' => $this->channel_id,
         ];
     }
+    
+    /**
+     * {@inheritdoc}
+     */
+    public function link(): string
```
davidcole1340, line 677:
```suggestion
    public function getLinkAttribute(): string
```
I think this would work better as an attribute, what do you think?

davidcole1340 (changes requested):

src/Discord/Parts/Channel/Message.php
valzargaming: Done

davidcole1340 (approved):

PR #546, opened 2021-06-09
Use Embed class when instantiating DiscordCommand

rachids:
Hello,

I was tinkering with my bot and discovered that when I type `!help` there are some default texts that is impossible to remove, since my bot is in french, I really wanted to be able to remove those defaults by applying my own description and title.

I also saw that it wasn't using the `Embed` class so I changed the old array for it.


davidcole1340:
Cheers!

PR #570, opened 2021-08-06
Quick getChannel fix when returning Thread

xWildFire:
Causing error when e.g $message->getChannel is returning a thread


src/Discord/Discord.php
```diff
@@ -1486,9 +1486,9 @@ public function __set(string $name, $value)
      *
      * @param string|int $channel_id Id of the channel.
      *
-     * @return Channel|null
+     * @return Channel|Thread|null
      */
-    public function getChannel($channel_id): ?Channel
+    public function getChannel($channel_id): Channel|Thread|null
```
davidcole1340, line 1491:
```suggestion
    public function getChannel($channel_id): ?Part
```

Union types were only introduced in PHP 8.0 but we need to support PHP 7.4.

davidcole1340 (changes requested):
Just one small change 😃

src/Discord/Discord.php
SQKo: I think I know what he mean.. in API docs it is mentioned that [Get Channel](https://discord.com/developers/docs/resources/channel#get-channel) may include a thread member object if channel is a thread.

```suggestion
public function getChannel($channel_id): ?Part
{
    foreach ($this->guilds as $guild) {
        foreach ($guild->channels as $channel) {
            if ($channel->id == $channel_id) {
                return $channel;
            } else if ($thread = $channel->threads->get('id', $channel_id)) {
                return $thread;
            }
        }
    }

    if ($channel = $this->private_channels->get('id', $channel_id)) {
        return $channel;
    }

    return null;
}
```
Not sure if the nested loop is overkill

SQKo:
I think I know what he mean.. in API docs it is mentioned that [Get Channel](https://discord.com/developers/docs/resources/channel#get-channel) may include a thread member object if channel is a thread.

```php
public function getChannel($channel_id): ?Part
{
    foreach ($this->guilds as $guild) {
        foreach ($guild->channels as $channel) {
            if ($channel->id == $channel_id) {
                return $channel;
            } else if ($thread = $channel->threads->get('id', $channel_id)) {
                return $thread;
            }
        }
    }

    if ($channel = $this->private_channels->get('id', $channel_id)) {
        return $channel;
    }

    return null;
}
```
Not sure if the nested loop is overkill

davidcole1340:
Yeah I think it's best if `getChannel` doesn't try to check for threads. Could add another `getThread` function.

PR #571, opened 2021-08-07
Allow old syntax of string argument in sendMessage [develop v7.0.0-rc0]

SQKo:
Upgrading from older version to **dev-develop** can be a pain when you have so many existing `sendMessage` codes that just simply send the message content string, and would be difficult to refactor to convert them all with `MessageBuilder`

### DIFF
--- a/src/Discord/Parts/Channel/Channel.php
+++ b/src/Discord/Parts/Channel/Channel.php
@@ -740,7 +740,7 @@ class Channel extends Part
     /**
      * Sends a message to the channel.
      *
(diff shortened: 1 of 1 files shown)


davidcole1340:
I've implemented it in [this commit](https://github.com/discord-php/DiscordPHP/commit/db9904cf45f13661c9ec1ac5cb2aea4b6c116071#diff-347cb634bd9dfce6baf790e117c398ae2c78812c67bfbbce182b3a32b4e1313fR757) which also adds all the other parameters which used to be there, thank you though :)

PR #575, opened 2021-08-14
Add magic __toString() in Channel.php

SQKo:
Add a magic string method to return formatted mention of text channel or the name of channel for other channel type

### DIFF
--- a/src/Discord/Parts/Channel/Channel.php
+++ b/src/Discord/Parts/Channel/Channel.php
@@ -979,9 +979,6 @@ class Channel extends Part
      */
     public function __toString(): string
     {
-        if ($this->type == self::TYPE_TEXT) {
-            return "<#{$this->id}>";
-        }
-        return $this->name;
+        return "<#{$this->id}>";
     }
 }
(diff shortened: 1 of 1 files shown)


src/Discord/Parts/Channel/Channel.php
```diff
@@ -971,4 +971,17 @@ public function getRepositoryAttributes(): array
...
     }
+    
+    /**
+     * Returns a formatted mention for text channel or name of the channel.
+     *
+     * @return string A formatted mention for text channel or name of the channel.
+     */
+    public function __toString(): string
+    {
+        if ($this->type == self::TYPE_TEXT) {
```
davidcole1340, line 982: Why only for text channels? Should work for all guild channels (text, voice, announcements etc).

SQKo: I wasn't sure about other types, Category was just giving plain text like `#Text Channel`, but yes voice works. I guess it should be fine for all types, I just tested that discord will automatically turn invalid channel id into "#deleted-channel"

PR #583, opened 2021-08-22
added customizeable custom_id feature

Ixiter:
Discord\Builders\Components\SelectMenu

Changed the constructor to expect a nullable param $customId. If the param is given, the custom_id is set to the given $customId. Otherwise its set with $this->generateUuid()
Added getter and setter for the custom_id to read and write them with the object instance.

### DIFF
--- a/src/Discord/Builders/Components/SelectMenu.php
+++ b/src/Discord/Builders/Components/SelectMenu.php
@@ -125,16 +125,6 @@ class SelectMenu extends Component
     }
 
     /**
-     * Creates a new select menu.
-     *
-     * @return self
-     */
-    public static function new(): self
-    {
-        return new self();
-    }
-
-    /**
      * Adds an option to the select menu. Maximum 25 options.
      *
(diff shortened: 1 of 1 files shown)


src/Discord/Builders/Components/SelectMenu.php
```diff
@@ -81,10 +81,47 @@ class SelectMenu extends Component
...
+    }
+
+    /**
+     * Creates a new select menu.
+     *
+     * @param string $customId The custom_id for this Selectmenu
+     * 
+     * @return self
+     */
+    public static function new($customId = null): self
```
SQKo, line 99: You are making duplicate of the `new` function instead of modifying existing one
https://github.com/discord-php/DiscordPHP/blob/2b91e2dae67e3c5541af01dc409e407607684872/src/Discord/Builders/Components/SelectMenu.php#L132

davidcole1340:
Thank you

PR #586, opened 2021-08-23
Add function to escape discord markdown

SQKo:
Used this function in my bot, useful to escape markdown and formatting symbols with backslashes

### DIFF
--- a/src/Discord/functions.php
+++ b/src/Discord/functions.php
@@ -243,3 +243,13 @@ function normalizePartId($id_field = 'id')
         return $part;
     };
 }
+
+/**
+ * Escape Discord markdown formatting
+ * 
+ * @return string the escaped string
+ */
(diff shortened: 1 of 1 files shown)


src/Discord/functions.php
```diff
@@ -243,3 +243,13 @@ function normalizePartId($id_field = 'id')
         return $part;
     };
 }
+
+/**
+ * Escape Discord markdown formatting
+ * 
+ * @return string the escaped string
+ */
+function EscapeMarkdown($text): string
```
davidcole1340, line 252:
```suggestion
function escapeMarkdown(string $text): string
```

davidcole1340 (changes requested):
Could you expand on what the function does in the comment? Also could you add the parameter type to the comment. Thank you!

davidcole1340:
Tnaks

PR #593, opened 2021-08-31
Documentation Fix example respondWithMessage

### LINKED ISSUE #578: Wrong usage in Docs - $interaction->respondWithMessage('Thanks')
Docs

**Describe the bug**
In the MessageBuilder Section, SelectMenu Example, you use $interaction->respondWithMessage('Thanks')
You cannot use a text as Message, you need a MessageBuilder.
Though .. would be cool to send simple text messages. 

**To Reproduce**
Just try the example

SQKo:
Replace respond message text with Message Builder... Fix #578


davidcole1340:
Thank you

PR #598, opened 2021-09-04
Fix URL typo getSplashAttribute() on Guild.php

SQKo:
Noticed this when looking around for "slash" where it is supposed to be "s**p**lash", I double checked the server that has invite background and it is correct:
https://cdn.discordapp.com/splashes/305281346455207936/0eb527e88251cab6145ecb8ee7b49bfa.jpg?size=1024


davidcole1340:
Thanks

PR #599, opened 2021-09-09
Fix PHP 8.1 deprecations

iggyvolz:
Mostly backwards-compatible PHP 8.1 deprecation fixes.  See https://wiki.php.net/rfc/internal_method_return_types and https://wiki.php.net/rfc/phase_out_serializable

On src/Discord/Helpers/Collection.php, `serialize()` serialized to a JSON string, which is not allowed by `__serialize` (which handles the serialization itself) so that is a (potential?) BC break.  From my understanding this is currently being used incorrectly anyways (you would end up getting a PHP-serialized JSON-encoded string on a call to `serialize()`) and the proper way to do that is JSONSerializable.

### DIFF
--- a/src/Discord/Helpers/Collection.php
+++ b/src/Discord/Helpers/Collection.php
@@ -22,7 +22,7 @@ use Traversable;
 /**
  * Collection of items. Inspired by Laravel Collections.
  */
-class Collection implements ArrayAccess, Serializable, JsonSerializable, IteratorAggregate, Countable
+class Collection implements ArrayAccess, JsonSerializable, IteratorAggregate, Countable
 {
     /**
      * The collection discriminator.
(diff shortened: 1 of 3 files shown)


davidcole1340:
Thanks for the PR!

Do you think you would be able to re-add the `serialize` and `unserialize` functions? I know that some users do use these outside of the interface implementation, so maybe we could just keep the functions while removing the `Serializable` interface as well as adding the `__serialize` and `__unserialize` functions?

iggyvolz:
Should I mark `serialize` and `unserialize` as `@deprecated` or just leave them as-is?

davidcole1340 (approved):
Should be fine as is. Thank you!

PR #602, opened 2021-09-18
Declare and Implement getFillableAttributes()

valzargaming:
There wasn't a way to get a part's $fillable array from outside of the class, which can make it tricky to dynamically create an object in certain use-cases. This resolve it by allowing the fillable property to be retrieved with a function. The main downside to this method is that the function has to be redeclared in every part that has a non-empty fillable array because of the usage of reflection, but I think that's a small price to pay for future maintainability.

### DIFF
--- a/src/Discord/Parts/Part.php
+++ b/src/Discord/Parts/Part.php
@@ -193,7 +193,7 @@ abstract class Part implements ArrayAccess, JsonSerializable
 	{
 		$fillable = array();
 		foreach (self::$fillable as $attr) {
-			if (!$context || in_array($context, $attrContexts)) {
+			if (!$context || in_array($context, self::$fillable)) {
 				$fillable[] = $attr;
 			}
 		}
--- a/src/Discord/Parts/Part.php
+++ b/src/Discord/Parts/Part.php
@@ -61,7 +61,7 @@ abstract class Part implements ArrayAccess, JsonSerializable
      *
      * @var array The array of attributes that can be mass-assigned.
      */
-    protected $fillable = [];
+    protected static $fillable = [];
 
     /**
      * The parts attributes.
@@ -178,11 +178,27 @@ abstract class Part implements ArrayAccess, JsonSerializable
     public function fill(array $attributes): void
     {
         foreach ($attributes as $key => $value) {
-            if (in_array($key, $this->fillable)) {
+            if (in_array($key, $this::getFillableAttributes())) {
                 $this->setAttribute($key, $value);
             }
         }
     }
+
+	/**
+     * Returns the fillable attributes.
+     *
+     * @return array
+     */
+    public static function getFillableAttributes($context = '')
+	{
+		$fillable = array();
+		foreach (self::$fillable as $attr) {
+			if (!$context || in_array($context, $attrContexts)) {
+				$fillable[] = $attr;
+			}
+		}
+		return $fillable;
+	}	
 
     /**
      * Checks if there is a mutator present.
@@ -246,7 +262,7 @@ abstract class Part implements ArrayAccess, JsonSerializable
             return;
         }
 
-        if (array_search($key, $this->fillable) !== false) {
+        if (array_search($key, $this::getFillableAttributes()) !== false) {
             $this->attributes[$key] = $value;
         }
     }
@@ -355,7 +371,7 @@ abstract class Part implements ArrayAccess, JsonSerializable
     {
         $data = [];
 
-        foreach (array_merge($this->fillable, $this->visible) as $key) {
+        foreach (array_merge($this::getFillableAttributes(), $this->visible) as $key) {
             if (in_array($key, $this->hidden)) {
                 continue;
             }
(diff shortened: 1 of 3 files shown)


src/Discord/Parts/Part.php
```diff
@@ -61,7 +61,7 @@ abstract class Part implements ArrayAccess, JsonSerializable
      *
      * @var array The array of attributes that can be mass-assigned.
      */
-    protected $fillable = [];
+    protected static $fillable = [];
```
davidcole1340, line 64: The children class need their `$fillable` changed to static as well otherwise PHP will error saying that you tried to redeclare as non-static. Saying that I think it should be `const` instead of `static` as fillables are immutable.

src/Discord/Parts/Part.php
```diff
@@ -178,12 +178,28 @@ public function fetch(): ExtendedPromiseInterface
...
 
+	/**
+     * Returns the fillable attributes.
+     *
+     * @return array
+     */
+    public static function getFillableAttributes($context = '')
+	{
+		$fillable = array();
+		foreach (self::$fillable as $attr) {
```
davidcole1340, line 195: `self::$fillable` will always reference `Part::$fillable` even if the function is called on a child class, you can use the `static::` keyword instead to reference the class that the function is called on (`static::$fillable`).

src/Discord/Parts/Part.php
```diff
@@ -178,12 +178,28 @@ public function fetch(): ExtendedPromiseInterface
...
             }
         }
     }
 
+	/**
+     * Returns the fillable attributes.
+     *
+     * @return array
+     */
+    public static function getFillableAttributes($context = '')
```
davidcole1340, line 192: I don't think the function should do any filtering, perhaps just return the `fillable` array and then it can be filtered later.

src/Discord/Parts/User/Client.php
```diff
@@ -42,7 +42,23 @@ class Client extends Part
...
      */
-    protected $fillable = ['id', 'username', 'email', 'verified', 'avatar', 'discriminator', 'bot', 'user', 'application'];
+    protected static $fillable = ['id', 'username', 'email', 'verified', 'avatar', 'discriminator', 'bot', 'user', 'application'];
+
+	/**
+     * Returns the fillable attributes.
+     *
+     * @return array
+     */
+    public static function getFillableAttributes($context = '')
```
davidcole1340, line 52: Not sure why the function is duplicated here? Also duplicated in `Application`.

valzargaming: Polymorphism, it overrides the function of the same name on the Part class

davidcole1340: Is it not the same function?

src/Discord/Parts/Part.php
SQKo: Just leaving my note here (since i posted in discord)

[Permissions.php](https://github.com/discord-php/DiscordPHP/blob/master/src/Discord/Parts/Permissions/Permission.php) Part has a mutable `$fillable`, unless modified, it cant be `const`
https://github.com/discord-php/DiscordPHP/blob/master/src/Discord/Parts/Permissions/Permission.php#L110-L111

valzargaming:
I don't remember if this is resolved yet, but if so then the following test function should work:
```php
function reflectionMachine($part = null, ?array $array_merge = [], ?string $message = '', ?string $command = '')//: string
{
	$tokens = array_merge($array_merge, explode(' ', $message));
	$method = new \ReflectionMethod('\\'.get_class($part), $command);
	$num = $method->getNumberOfParameters();
	$tokens = array_slice($tokens, 0, $num);
	if ($part instanceof Part && count($tokens) < $num) { //Too few parameters passed to function
		echo "num: $num" . PHP_EOL; echo 'token count: ' . count($tokens) . PHP_EOL;
		$parameters = [];
		for ($x=0;$x<$num;$x++) {
			$parameters[] = new ReflectionParameter([$part, $command], $x);
		}
		$return = "The `$command` command requires `$num` parameters: ";
		foreach ($parameters as $parameter) {
			$return .= "`{$parameter->getName()}`, ";
		}
		$return = substr($return, 0, strlen($return)-2) . '.';
		return $return;
	}
	return call_user_func_array(array($part, $command), $tokens);
}
```

src/Discord/Parts/Part.php
valzargaming: I did some testing and confirmed that this is indeed the case; Unless the function itself is overridden in the child classes, self::$fillable will call the parent method if that's where the function is being defined. Each individual Part would need its own getFillableAttribute function if the self keyword was to be used.

src/Discord/Parts/User/Client.php
valzargaming: This function currently uses the self keyword to refer to $fillable, so this function would get the fillable attributes of the parent class Part, which is incorrect behavior.

PR #606, opened 2021-10-03
Avoid hexdec deprecated warning (PHP 7.4+)

jocafamaka:
Avoid deprecated warning, and possible future error:
```
Deprecated: Invalid characters passed for attempted conversion, these have been ignored in ***/vendor/team-reflex/discord-php/src/Discord/Parts/Embed/Embed.php on line 484
```

This happens when the color is passed as:
```PHP
$embed = new Embed($discord, [
        "type" => Embed::TYPE_RICH,
        "description" => $desc,
        "color" => '#2c2c2c'
    ]);
```

Simple sanitization performed before parse to conversion.

### DIFF
--- a/src/Discord/Parts/Embed/Embed.php
+++ b/src/Discord/Parts/Embed/Embed.php
@@ -481,7 +481,7 @@ class Embed extends Part
         }
 
         if (! is_array($color)) {
-            return hexdec(((string) $color));
+            return hexdec((preg_replace('/[^0-9a-fA-FxX]/', '', (string) $color)));
         }
 
         if (count($color) < 1) {
--- a/src/Discord/Parts/Embed/Embed.php
+++ b/src/Discord/Parts/Embed/Embed.php
@@ -481,7 +481,7 @@ class Embed extends Part
         }
 
         if (! is_array($color)) {
-            return hexdec((preg_replace('/[^0-9a-fA-FxX]/', '', (string) $color)));
+            return hexdec((str_replace('#', '', (string) $color)));
         }
 
         if (count($color) < 1) {


SQKo:
I don't see why it is necessary, in PHP passing hex color should be just this way:
```php
'color' => '2c2c2c'
```
or
```php
'color' => 0x2c2c2c;
```
or
```php
'color' => 0x2c2c2c;
```
And those are valid with hexdec in php 7.4

Also Discord API treat the colors with int values, I think that doing '#2c2c2c' is user specific behavior, not caused or limited by API or even the language itself, so conversion should be done manually by the user.

jocafamaka:
I get that, but if it's something that might break the lib in the future, why not treat it? It's a harmless change.

The vast majority of users (myself included) use decimal colors this way, largely because it's standard in CSS.

And even other parts of the code imply that this is valid:
https://github.com/discord-php/DiscordPHP/blob/c635bcb5bd554284f7e77f088b23081ea62a1edd/src/Discord/functions.php#L95-L113

And the deprecated message can be quite annoying, in addition to becoming a fatal error in future versions of PHP.

SQKo:
This library is made for php 7.4+ so avoiding the errors suggested is a bad move

the getColor function provided in functions.php is a helper/support function for user to use, it is not used anywhere in the library.

If the goal is only to allow CSS format color code, then a simple check and remove for the first '#' is enough than using regex to replace all invalid characters which is overkill

Even worse, if the converted color code is invalid, Discord API will silently not show the Embed.

jocafamaka:
> If the goal is only to allow CSS format color code, then a simple check and remove for the first '#' is enough than using regex to replace all invalid characters which is overkill

So be it, my goal is just to make things easier!

> Even worse, if the converted color code is invalid, Discord API will silently not show the Embed.

You say this as if this change would allow this behavior, which is already what happens, clearly if the hex code is wrong it won't work.

src/Discord/Parts/Embed/Embed.php
```diff
@@ -481,7 +481,7 @@ protected static function resolveColor($color)
         }
 
         if (! is_array($color)) {
-            return hexdec(((string) $color));
+            return hexdec((preg_replace('/[^0-9a-fA-FxX]/', '', (string) $color)));
```
SQKo, line 484: replace with
```php
return hexdec((str_replace('#', '', (string) $color)));
```

SQKo (changes requested):
This should be enough i guess?

jocafamaka:
>This should be enough i guess?

I believe so!

davidcole1340 (approved):
Thanks!

PR #608, opened 2021-10-05
Add banner and accent_color to User.php

SQKo:
Add user banner (url and hash) and accent_color attributes to User part, additionally add missing GIF format in avatar url


src/Discord/Parts/User/User.php
```diff
@@ -151,6 +154,37 @@ protected function getAvatarHashAttribute(): string
...
     }
 
+        /**
+     * Returns the banner URL for the client.
+     *
+     * @param string $format The image format.
+     * @param int    $size   The size of the image.
+     *
+     * @return string|null The URL to the clients banner.
+     */
```
davidcole1340, line 164:
```suggestion
    /**
     * Returns the banner URL for the client.
     *
     * @param string $format The image format.
     * @param int    $size   The size of the image.
     *
     * @return string|null The URL to the clients banner.
     */
```

davidcole1340:
Thanks

PR #622, opened 2021-11-17
getAvatarAttribute throws fatal error if null

valzargaming:


src/Discord/Parts/User/Member.php
```diff
@@ -402,7 +402,7 @@ protected function getJoinedAtAttribute(): ?Carbon
     public function getAvatarAttribute(string $format = 'jpg', int $size = 1024): string
```
davidcole1340, line 402:
```suggestion
    public function getAvatarAttribute(string $format = 'jpg', int $size = 1024): ?string
```

Should change the type hint instead

src/Discord/Parts/User/Member.php
```diff
@@ -402,7 +402,7 @@ protected function getJoinedAtAttribute(): ?Carbon
...
     {
         if (is_null($this->attributes['avatar'])) {
-            return null;
+            return '';
```
davidcole1340, line 405:
```suggestion
            return null;
```

As above

davidcole1340 (changes requested):

PR #635, opened 2021-12-10
Add missing properties in Guild.php

SQKo:
property `premium_progress_bar_enabled` is also updatable

Ref: https://github.com/discord/discord-api-docs/pull/4001


SQKo:
Re-checked the documentation and apparently `welcome_screen` and `nsfw_level` is also missing

PR #637, opened 2021-12-11
Add Welcome Screen Channels in Guild.php

SQKo:
Add the missing [welcome screen](https://discord.com/developers/docs/resources/guild#welcome-screen-object) field in Guild, as well the methods to get & update it.

- [ ] Add documentation

### DIFF
--- a/src/Discord/Parts/Guild/Guild.php
+++ b/src/Discord/Parts/Guild/Guild.php
@@ -665,7 +665,7 @@ class Guild extends Part
     {
         return $this->http->get(Endpoint::bind(Endpoint::GUILD_WELCOME_SCREEN, $this->id))->then(function ($response) {
             $welcome_screen = $this->discord->factory(WelcomeScreen::class, $response, true);
-            $this->welcome_screen = $welcome_screen;
+            $this->attributes['welcome_screen'] = $welcome_screen;
 
             return $welcome_screen;
         });
@@ -697,7 +697,7 @@ class Guild extends Part
 
         return $this->http->patch(Endpoint::bind(Endpoint::GUILD_WELCOME_SCREEN, $this->id), $options)->then(function ($response) {
             $welcome_screen = $this->discord->factory(WelcomeScreen::class, $response, true);
-            $this->welcome_screen = $welcome_screen;
+            $this->attributes['welcome_screen'] = $welcome_screen;
 
             return $welcome_screen;
         });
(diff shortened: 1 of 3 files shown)


src/Discord/Parts/Guild/Guild.php
```diff
@@ -165,6 +166,7 @@ class Guild extends Part
         'max_video_channel_users',
         'approximate_member_count',
         'approximate_presence_count',
+        'welcome_screen',
```
davidcole1340, line 169: There needs to be a `getWelcomeScreenAttribute()` if the `welcome_screen` attribute is to return an instance of `WelcomeScreen`

src/Discord/Parts/Guild/Guild.php
```diff
@@ -655,6 +657,53 @@ public function searchMembers(array $options): ExtendedPromiseInterface
         });
     }
 
+    /**
+     * Returns the Welcome Screen object for the guild.
+     *
+     * @return ExtendedPromiseInterface
+     */
+    public function getWelcomeScreen(): ExtendedPromiseInterface
```
davidcole1340, line 665: Since the welcome screen is only present when sent down inside an invite, maybe we should 'hide' the `welcome_screen` attribute and force the users to use `getWelcomeScreen()`.

e.g. remove `welcome_screen` from the phpdoc (but leave it in fillable), then in this function check `isset($this->attributes['welcome_screen'])` and return that if so

src/Discord/Parts/Guild/WelcomeScreen.php
```diff
@@ -0,0 +1,30 @@
...
+
+use Discord\Parts\Part;
+
+/**
+ * A Welcome Screen of a Guild
+ *
+ * @link https://discord.com/developers/docs/resources/guild#welcome-screen-object-welcome-screen-structure
+ * 
+ * @property string                      $description      The server description shown in the welcome screen.
+ * @property Collection|WelcomeChannel[] $welcome_channels The channels shown in the welcome screen, up to 5.
```
davidcole1340, line 22: Need a `getWelcomeChannels()` function to return a collection and/or an array of parts

davidcole1340 (dismissed):

src/Discord/Parts/Guild/Guild.php
SQKo: The `welcome_screen` may be updated,  it's not only present with invites.

I think to make it like the fetch() function:
```suggestion
    public function getWelcomeScreen(bool $fresh = false): ExtendedPromiseInterface
    {
        if (! $fresh && isset($this->attributes['welcome_screen'])) {
            return \React\Promise\resolve($this->attributes['welcome_screen']);
        }

```

valzargaming (approved):

Issue #653, opened 2021-12-23
TODO: Add X-Audit-Log-Reason where possible


SQKo:
There are some endpoints that accepts `X-Audit-Log-Reason` header as a custom Audit Log reason, e.g. the [Create Guild Ban](https://discord.com/developers/docs/resources/guild#create-guild-ban), since Discord is deprecating the `reason` from the query parameter. Probably need to implement the logic from [DiscordPHP-Http](https://github.com/discord-php/DiscordPHP-Http)

Most of the endpoints that support this are with the repository save() with POST/PATCH or delete() with DELETE 

Should we make the logic in the `save()` and `delete()` methods? Although that's covering the whole endpoints in general 
e.g. `->save($part, $reason)`  then we could add in each part/repository to mark which create/update supports X-Audit-Log-Reason or not

Any better way to implement this?

PR #656, opened 2021-12-26
Add reason parameter to some endpoint methods

SQKo:
Add an optional `$reason` parameter as `X-Audit-Log-Reason` header in some methods that support it (i.e. http `->post()` `->patch()` `->put()` and `->delete()`).
This does not cover audit log endpoints from create/update repository.

#653 

~~Note: not really tested yet.~~

- [ ] Update documentation

### DIFF
--- a/src/Discord/Parts/Channel/Channel.php
+++ b/src/Discord/Parts/Channel/Channel.php
@@ -492,7 +492,7 @@ class Channel extends Part
      * Bulk deletes an array of messages.
      *
      * @param array|Traversable $messages An array of messages to delete.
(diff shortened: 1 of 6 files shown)


valzargaming (approved):

SQKo:
Fixed missing `$reason` parameter in moveMember()

Removed audit-log `$reason` from delete single message & update role positions as I confirmed to Discord unofficial API discord server that it does not work

However bulk messages & unpin message will not render the reason in audit log which devsnek said will be fixed soon, so it's kept.

valzargaming (approved):

PR #659, opened 2021-12-29
Need examples :)

yuraMovsesyan:
it's so hard to find any information about Discord PHP. And in order to make life easier for the developer, you need to add examples of simple code.

so, for example, the classic ping-pong implementation was added :D

### DIFF
--- a/examples/ping.php
+++ b/examples/ping.php
@@ -1,25 +1,37 @@
 <?php
+
+/**
+ * Example Bot with Discord-PHP
+ *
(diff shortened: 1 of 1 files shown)


SQKo:
Did you intend to add example with the Browser class? otherwise It's not necessary.

yuraMovsesyan:
Chet I overlooked this moment...

can add this code as an example, which you wrote

SQKo (approved):

PR #660, opened 2021-12-30
Fix 32 bit php compatibility with permission bits

SQKo:
Added a new composer dependency https://github.com/brick/math
Added a new helper class **Bitwise.php**
Added missing permissions `manage_events` & `moderate_members`

Changed the permission constants back to hexadecimal form as doing bit shift left on bits above 31 on 32 bit php will give unexpected result due to integer overflow.
Note: this is only safe up to 44th bits, discord permission beyond 1 << 43 would require the constant value wrapped as numeric string (i.e. `1 << 44` = `'17592186044416'`)

Potentially breaking changes:

`$role->bitwise` now returns string instead of int for 32 bit PHP.

Fixed `Member::getPermissions()` returning all true when a member has permission with bits above 31 set.

Might need to be deeply tested (e.g. with combination of multiple permissions of roles and channels). 64 bit PHP should not be affected.

This fix is also applicable for v6.x since gateway API v8 is also affected.

Reference:
https://discord.com/developers/docs/topics/permissions#permissions-bitwise-permission-flags

### Quick Test
Tested on the bot running DiscordPHP
`@everyone` has no permissions
The bot role has "View Channels" `view_channel`  (`1 << 10`) & "Use Application Commands" `use_slash_commands` (`1 << 31`) permission set

```php
$botmember = $message->channel->guild->members->offsetGet($discord->id);
var_dump($botmember);
$botperms = $botmember->getPermissions();
var_dump($botperms);
```

All irrelevant var_dump data is removed from the output below

Before:
```
object(Discord\Parts\Permissions\RolePermission)#1822 (41) {
  ["create_instant_invite"]=>
  bool(true)
  ["manage_channels"]=>
  bool(true)
  ["view_channel"]=>
  bool(true)
  ["manage_roles"]=>
  bool(true)
  ["manage_webhooks"]=>
  bool(true)
  ["add_reactions"]=>
  bool(true)
  ["send_messages"]=>
  bool(true)
  ["send_tts_messages"]=>
  bool(true)
  ["manage_messages"]=>
  bool(true)
  ["embed_links"]=>
  bool(true)
  ["attach_files"]=>
  bool(true)
  ["read_message_history"]=>
  bool(true)
  ["mention_everyone"]=>
  bool(true)
  ["use_external_emojis"]=>
  bool(true)
  ["use_slash_commands"]=>
  bool(false)
  ["manage_threads"]=>
  bool(true)
  ["use_public_threads"]=>
  bool(true)
  ["use_private_threads"]=>
  bool(true)
  ["use_external_stickers"]=>
  bool(true)
  ["send_messages_in_threads"]=>
  bool(true)
  ["priority_speaker"]=>
  bool(true)
  ["stream"]=>
  bool(true)
  ["connect"]=>
  bool(true)
  ["speak"]=>
  bool(true)
  ["mute_members"]=>
  bool(true)
  ["deafen_members"]=>
  bool(true)
  ["move_members"]=>
  bool(true)
  ["use_vad"]=>
  bool(true)
  ["request_to_speak"]=>
  bool(true)
  ["manage_events"]=>
  bool(true)
  ["start_embedded_activities"]=>
  bool(true)
  ["kick_members"]=>
  bool(true)
  ["ban_members"]=>
  bool(true)
  ["administrator"]=>
  bool(true)
  ["manage_guild"]=>
  bool(true)
  ["view_audit_log"]=>
  bool(true)
  ["view_guild_insights"]=>
  bool(true)
  ["change_nickname"]=>
  bool(true)
  ["manage_nicknames"]=>
  bool(true)
  ["manage_emojis"]=>
  bool(true)
  ["bitwise"]=>
  int(2147483647)
}

object(Discord\Parts\Permissions\RolePermission)#1870 (41) {
  ["create_instant_invite"]=>
  bool(true)
  ["manage_channels"]=>
  bool(true)
  ["view_channel"]=>
  bool(true)
  ["manage_roles"]=>
  bool(true)
  ["manage_webhooks"]=>
  bool(true)
  ["add_reactions"]=>
  bool(true)
  ["send_messages"]=>
  bool(true)
  ["send_tts_messages"]=>
  bool(true)
  ["manage_messages"]=>
  bool(true)
  ["embed_links"]=>
  bool(true)
  ["attach_files"]=>
  bool(true)
  ["read_message_history"]=>
  bool(true)
  ["mention_everyone"]=>
  bool(true)
  ["use_external_emojis"]=>
  bool(true)
  ["use_slash_commands"]=>
  bool(true)
  ["manage_threads"]=>
  bool(true)
  ["use_public_threads"]=>
  bool(true)
  ["use_private_threads"]=>
  bool(true)
  ["use_external_stickers"]=>
  bool(true)
  ["send_messages_in_threads"]=>
  bool(true)
  ["priority_speaker"]=>
  bool(true)
  ["stream"]=>
  bool(true)
  ["connect"]=>
  bool(true)
  ["speak"]=>
  bool(true)
  ["mute_members"]=>
  bool(true)
  ["deafen_members"]=>
  bool(true)
  ["move_members"]=>
  bool(true)
  ["use_vad"]=>
  bool(true)
  ["request_to_speak"]=>
  bool(true)
  ["manage_events"]=>
  bool(true)
  ["start_embedded_activities"]=>
  bool(true)
  ["kick_members"]=>
  bool(true)
  ["ban_members"]=>
  bool(true)
  ["administrator"]=>
  bool(true)
  ["manage_guild"]=>
  bool(true)
  ["view_audit_log"]=>
  bool(true)
  ["view_guild_insights"]=>
  bool(true)
  ["change_nickname"]=>
  bool(true)
  ["manage_nicknames"]=>
  bool(true)
  ["manage_emojis"]=>
  bool(true)
  ["bitwise"]=>
  int(-1)
}
```

After:
```
object(Discord\Parts\Permissions\RolePermission)#1822 (42) {
  ["create_instant_invite"]=>
  bool(false)
  ["manage_channels"]=>
  bool(false)
  ["view_channel"]=>
  bool(true)
  ["manage_roles"]=>
  bool(false)
  ["manage_webhooks"]=>
  bool(false)
  ["add_reactions"]=>
  bool(false)
  ["send_messages"]=>
  bool(false)
  ["send_tts_messages"]=>
  bool(false)
  ["manage_messages"]=>
  bool(false)
  ["embed_links"]=>
  bool(false)
  ["attach_files"]=>
  bool(false)
  ["read_message_history"]=>
  bool(false)
  ["mention_everyone"]=>
  bool(false)
  ["use_external_emojis"]=>
  bool(false)
  ["use_slash_commands"]=>
  bool(true)
  ["manage_threads"]=>
  bool(false)
  ["use_public_threads"]=>
  bool(false)
  ["use_private_threads"]=>
  bool(false)
  ["use_external_stickers"]=>
  bool(false)
  ["send_messages_in_threads"]=>
  bool(false)
  ["priority_speaker"]=>
  bool(false)
  ["stream"]=>
  bool(false)
  ["connect"]=>
  bool(false)
  ["speak"]=>
  bool(false)
  ["mute_members"]=>
  bool(false)
  ["deafen_members"]=>
  bool(false)
  ["move_members"]=>
  bool(false)
  ["use_vad"]=>
  bool(false)
  ["request_to_speak"]=>
  bool(false)
  ["manage_events"]=>
  bool(false)
  ["start_embedded_activities"]=>
  bool(false)
  ["kick_members"]=>
  bool(false)
  ["ban_members"]=>
  bool(false)
  ["administrator"]=>
  bool(false)
  ["manage_guild"]=>
  bool(false)
  ["view_audit_log"]=>
  bool(false)
  ["view_guild_insights"]=>
  bool(false)
  ["change_nickname"]=>
  bool(false)
  ["manage_nicknames"]=>
  bool(false)
  ["manage_emojis"]=>
  bool(false)
  ["moderate_members"]=>
  bool(false)
  ["bitwise"]=>
  string(10) "2147484672"
}

object(Discord\Parts\Permissions\RolePermission)#1870 (42) {
  ["create_instant_invite"]=>
  bool(false)
  ["manage_channels"]=>
  bool(false)
  ["view_channel"]=>
  bool(true)
  ["manage_roles"]=>
  bool(false)
  ["manage_webhooks"]=>
  bool(false)
  ["add_reactions"]=>
  bool(false)
  ["send_messages"]=>
  bool(false)
  ["send_tts_messages"]=>
  bool(false)
  ["manage_messages"]=>
  bool(false)
  ["embed_links"]=>
  bool(false)
  ["attach_files"]=>
  bool(false)
  ["read_message_history"]=>
  bool(false)
  ["mention_everyone"]=>
  bool(false)
  ["use_external_emojis"]=>
  bool(false)
  ["use_slash_commands"]=>
  bool(true)
  ["manage_threads"]=>
  bool(false)
  ["use_public_threads"]=>
  bool(false)
  ["use_private_threads"]=>
  bool(false)
  ["use_external_stickers"]=>
  bool(false)
  ["send_messages_in_threads"]=>
  bool(false)
  ["priority_speaker"]=>
  bool(false)
  ["stream"]=>
  bool(false)
  ["connect"]=>
  bool(false)
  ["speak"]=>
  bool(false)
  ["mute_members"]=>
  bool(false)
  ["deafen_members"]=>
  bool(false)
  ["move_members"]=>
  bool(false)
  ["use_vad"]=>
  bool(false)
  ["request_to_speak"]=>
  bool(false)
  ["manage_events"]=>
  bool(false)
  ["start_embedded_activities"]=>
  bool(false)
  ["kick_members"]=>
  bool(false)
  ["ban_members"]=>
  bool(false)
  ["administrator"]=>
  bool(false)
  ["manage_guild"]=>
  bool(false)
  ["view_audit_log"]=>
  bool(false)
  ["view_guild_insights"]=>
  bool(false)
  ["change_nickname"]=>
  bool(false)
  ["manage_nicknames"]=>
  bool(false)
  ["manage_emojis"]=>
  bool(false)
  ["moderate_members"]=>
  bool(false)
  ["bitwise"]=>
  string(10) "2147484672"
}
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,8 @@
         "ext-zlib": "*",
         "discord-php/http": "^9.0.5",
         "react/child-process": "^0.6.2",
-        "discord/interactions": "^2"
+        "discord/interactions": "^2",
+        "brick/math": "^0.9.3"
     },
     "require-dev": {
         "symfony/var-dumper": "*",
--- a/src/Discord/Helpers/Bitwise.php
+++ b/src/Discord/Helpers/Bitwise.php
@@ -1,0 +1,86 @@
+<?php
+
+/*
+ * This file is a part of the DiscordPHP project.
+ *
+ * Copyright (c) 2015-present David Cole <david.cole1340@gmail.com>
+ *
+ * This file is subject to the MIT license that is bundled
+ * with this source code in the LICENSE.md file.
+ */
+
+namespace Discord\Helpers;
+
+use Brick\Math\BigInteger;
+
+/**
+ * A compat helper to handle bitwise operation in 32 bit php wrapping BigInteger
+ */
+class Bitwise
+{
+    /**
+     * @param BigInteger|int|float|string $a
+     * @param BigInteger|int|float|string $b
+     *
+     * @return BigInteger|int $a & $b
+     */
+    public static function and($a, $b)
+    {
+        return (PHP_INT_SIZE == 4) ? BigInteger::of($a)->and($b) : $a & $b;
+    }
+
+    /**
+     * @param BigInteger|int|float|string $a
+     * @param BigInteger|int|float|string $b
+     *
+     * @return BigInteger|int $a | $b
+     */
+    public static function or($a, $b)
+    {
+        return (PHP_INT_SIZE == 4) ? BigInteger::of($a)->or($b) : $a | $b;
+
+    }
+
+    /**
+     * @param BigInteger|int|float|string $a
+     * @param BigInteger|int|float|string $b
+     *
+     * @return BigInteger|int $a ^ $b
+     */
+    public static function xor($a, $b)
+    {
+        return (PHP_INT_SIZE == 4) ? BigInteger::of($a)->xor($b) : $a ^ $b;
+    }
+
+    /**
+     * @param BigInteger|int|float|string $a
+     *
+     * @return BigInteger|int ~ $a
+     */
+    public static function not($a)
+    {
+        return (PHP_INT_SIZE == 4) ? BigInteger::of($a)->not() : ~$a;
+    }
+
+    /**
+     * @param BigInteger|int|float|string $a
+     * @param BigInteger|int|float|string $b
+     *
+     * @return BigInteger|int $a << $b
+     */
+    public static function shiftLeft($a, $b)
+    {
+        return (PHP_INT_SIZE == 4) ? BigInteger::of($a)->shiftedLeft($b) : $a << $b;
+    }
+
+    /**
+     * @param BigInteger|int|float|string $a
+     * @param BigInteger|int|float|string $b
+     *
+     * @return BigInteger|int $a >> $b
+     */
+    public static function shiftRight($a, $b)
+    {
+        return (PHP_INT_SIZE == 4) ? BigInteger::of($a)->shiftedRight($b) : $a >> $b;
+    }
+}
--- a/src/Discord/Parts/Permissions/Permission.php
+++ b/src/Discord/Parts/Permissions/Permission.php
@@ -11,38 +11,42 @@
 
 namespace Discord\Parts\Permissions;
 
+use Brick\Math\BigInteger;
 use Discord\Discord;
+use Discord\Helpers\Bitwise;
 use Discord\Parts\Part;
 
 /**
  * Permission represents a set of permissions for a given role or overwrite.
  *
- * @property int  $bitwise
- * @property bool $create_instant_invite
- * @property bool $manage_channels
- * @property bool $view_channel
- * @property bool $manage_roles
- * @property bool $manage_webhooks
+ * @property int|string $bitwise
+ * @property bool       $create_instant_invite
+ * @property bool       $manage_channels
+ * @property bool       $view_channel
+ * @property bool       $manage_roles
+ * @property bool       $manage_webhooks
  */
 abstract class Permission extends Part
 {
+    // Note: Bits above 44th must use numeric string for 32 bit compatibility
+
     /**
      * Array of permissions that only apply to voice channels.
      *
      * @var array
      */
     public const VOICE_PERMISSIONS = [
-        'priority_speaker' => (1 << 8),
-        'stream' => (1 << 9),
-        'connect' => (1 << 20),
-        'speak' => (1 << 21),
-        'mute_members' => (1 << 22),
-        'deafen_members' => (1 << 23),
-        'move_members' => (1 << 24),
-        'use_vad' => (1 << 25),
-        'request_to_speak' => (1 << 32),
-        'manage_events' => (1 << 33),
-        'start_embedded_activities' => (1 << 39),
+        'priority_speaker' => 0x100,
+        'stream' => 0x200,
+        'connect' => 0x100000,
+        'speak' => 0x200000,
+        'mute_members' => 0x400000,
+        'deafen_members' => 0x800000,
+        'move_members' => 0x1000000,
+        'use_vad' => 0x2000000,
+        'request_to_speak' => 0x100000000,
+        'manage_events' => 0x200000000,
+        'start_embedded_activities' => 0x8000000000,
     ];
 
     /**
@@ -51,21 +55,21 @@ abstract class Permission extends Part
      * @var array
      */
     public const TEXT_PERMISSIONS = [
-        'add_reactions' => (1 << 6),
-        'send_messages' => (1 << 11),
-        'send_tts_messages' => (1 << 12),
-        'manage_messages' => (1 << 13),
-        'embed_links' => (1 << 14),
-        'attach_files' => (1 << 15),
-        'read_message_history' => (1 << 16),
-        'mention_everyone' => (1 << 17),
-        'use_external_emojis' => (1 << 18),
-        'use_slash_commands' => (1 << 31),
-        'manage_threads' => (1 << 34),
-        'use_public_threads' => (1 << 35),
-        'use_private_threads' => (1 << 36),
-        'use_external_stickers' => (1 << 37),
-        'send_messages_in_threads' => (1 << 38),
+        'add_reactions' => 0x40,
+        'send_messages' => 0x800,
+        'send_tts_messages' => 0x1000,
+        'manage_messages' => 0x2000,
+        'embed_links' => 0x4000,
+        'attach_files' => 0x8000,
+        'read_message_history' => 0x10000,
+        'mention_everyone' => 0x20000,
+        'use_external_emojis' => 0x40000,
+        'use_slash_commands' => 0x80000000,
+        'manage_threads' => 0x400000000,
+        'use_public_threads' => 0x800000000,
+        'use_private_threads' => 0x1000000000,
+        'use_external_stickers' => 0x2000000000,
+        'send_messages_in_threads' => 0x4000000000,
     ];
 
     /**
@@ -74,15 +78,17 @@ abstract class Permission extends Part
      * @var array
      */
     public const ROLE_PERMISSIONS = [
-        'kick_members' => (1 << 1),
-        'ban_members' => (1 << 2),
-        'administrator' => (1 << 3),
-        'manage_guild' => (1 << 5),
-        'view_audit_log' => (1 << 7),
-        'view_guild_insights' => (1 << 19),
-        'change_nickname' => (1 << 26),
-        'manage_nicknames' => (1 << 27),
-        'manage_emojis' => (1 << 30),
+        'kick_members' => 0x2,
+        'ban_members' => 0x4,
+        'administrator' => 0x8,
+        'manage_guild' => 0x20,
+        'view_audit_log' => 0x80,
+        'view_guild_insights' => 0x80000,
+        'change_nickname' => 0x4000000,
+        'manage_nicknames' => 0x8000000,
+        'manage_emojis' => 0x40000000,
+        'manage_events' => 0x200000000,
+        'moderate_members' => 0x10000000000,
     ];
 
     /**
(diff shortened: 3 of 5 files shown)


valzargaming (approved):

shehi:
I strongly advise against this. It will complicate development very much. 32-bit is history, almost on all fronts. Instead of moving forward, why are we stickying ourselves to this archaic and heavily inconvenient way of development? Strings are error-prone when your data's nature is int-based and we will have a lot bugs caused by these. Just drop 32-bit support as countless other libs do and move on please? Pretty please?

SQKo:
I'm just trying to help sir, this bug apparently also affected v6.0.0 since it uses v8 API Gateway, after 6 months release, wouldn't it be a serious issue?

So I hope this get patched with this PR as well.

Someone who have been using this library and did not notice they are facing the bug wont be happy because nobody here ever said it does not support 32 bit PHP.

https://discord.com/developers/docs/topics/permissions
The official docs explained here the permission data is serialized as string anyway:
> Permissions are stored within a variable-length integer serialized into a string, and are calculated using bitwise operations. For example, the permission value `123` will be serialized as `"123"`

And it's written to use BigInteger libraries, so I just did follow the docs:
>  we recommend deserializing the permissions using your languages' Big Integer libraries

It's only the permissions part affected, not the whole library, going to v6.0.4 just to throw error the library can't be used in 32 bit is just silly.

To be also note that Discord even send snowflake in serialized string, and this library does that, without changing it to integer, reflecting the docs. There is no one here being inconvenient with the way of the development, unless Discord did send them as 64 bit integer at first.

> And changing all those permission bits from `1 << x` to God-awful, unreadable `0xblahblah` is the worst gift we can get for the New Year! :D

No one but a techie enough would bother with readability of that bits in the source code. Bot developers are just going to use the mapped array keys as the permission names. Use `define` if that'd still bother someone, PHP is not a compiled language, just take it easy.

> P.S.: And why don't we use php-bcmath with countless years of battle testing and instead come up with this library (brick something) which doesn't even have proper v1.0?!

The library [brick/math uses 0.x.y versioning](https://github.com/brick/math#project-status--release-process) where x is the breaking change version:
> The current releases are numbered `0.x.y`. When a non-breaking change is introduced (adding new methods, optimizing existing code, etc.), `y` is incremented.

The [library also support](https://github.com/brick/math#requirements) both php-bcmath and php-gmp extension that will be automatically used.
> Although the library can work seamlessly on any PHP installation, it is highly recommended that you install the GMP or BCMath extension to speed up calculations. The fastest available calculator implementation will be automatically selected at runtime.

Now if there is someone with 32 bit php and worrying about string issue, they can just solve it by enabling GMP/BCMath extension.

Being not required to use the extension, simply means this PR is not a breaking change for a patch.

And you do not need to worry if you use php 64 bit, nothing changed due to this PR. That's why we selected that library to wrap it easy, as a hand to the 32 bit php.
All will be fine, everyone would be happy. Let me know if this patch is causing your bot a trouble.

> Just drop 32-bit support as countless other libs do and move on please? Pretty please? Just put one big exception on lib constructor to detect 32-bit and throw and be done with it. We are almost in 2022 - 64bit itself is already 2000s.

Well I'm sorry to say this personally, yes I still run 32 bit on my laptop for development (No, I run my bot at 64 bit php on production server), pardon me living in 3rd world countries. But this library helped me a lot. And I'm just trying to help it back. Otherwise I wouldn't be here from start.

Regardless this gets added or not, or otherwise drop support for 64 bit, that's fine, I can still fork this and use it personally for my development.

shehi:
Please don't make us suffer just because you don't want to put 64bit PHP on
your local dev. Whoever is in charge of this project, should stop this PR,
get rid of 32bit support and move on. Just do major version release if
necessary, drop 32bit there.

This project barely has proper test coverage, but instead we are letting
this through. If this gets in, I personally will abandon bot development in
PHP and switch to Js. This is a mess I don't want in a serious
Discord-heavy project where permissions are extremely crucial (I am
building guildplanner.pro which runs on the back of Discord).

shehi:
Sorry I am outside that's why posting in bits.

Few more concerns:

1. We shouldn't have multiple variable precision libraries in one lib. Use bcmath, drop anything else. Noone needs polyfill for a widely known extension.

2. This merge will complicate development, and it make readable code more complex. That's a big minus for a project which is maintained by a handful few. 

I will read your reply thoroughly in 30mins and see if I haven't answered any points there.

valzargaming:
Thank you for the valuable feedback @shehi. To address your original concern with supporting 32-bit, DiscordPHP already does so and has for some time (see: [getSnowflakeTimestamp()](https://github.com/discord-php/DiscordPHP/blob/74fbd0efec4d66fcf4d99ac1346eb8076f0b0957/src/Discord/functions.php#L203)) and expanding its support for it only makes sense at this point.

Regarding using external libraries as dependencies to support 32-bit precision I have to concede on your opinion that it is a bit of extra work for us maintainers to make sure it works correctly on those systems, however we also acknowledge there are valid use cases where an OS may be 32-bit, particularly those that are designed to be cheap and run with a limited amount of memory available, thus we value this lib's ability to be compatible with a 32-bit PHP installation. This leaves us with 5 choices right now:
1) Use no other dependency at all, maintaining bit conversion methods ourselves within the library using methods such as with base_convert()
2) Use brick/math, which is also MIT licensed, and even the author of the library recommends just using GMP
3) Use the GMP extension (discordphp/slash already requires this so there's no issue there)
4) Update existing all existing 32-bit supported functions to default to 64-bit and require an additional dependency to process those functions if PHP_INT_SIZE === 4 during composer require, or
5) Remove all existing 32-bit support in the library and only support 64-bit, thus breaking at least a few known applications

We already have prepared working solutions for options 1 and 2. Option 3 would be the most ideal after those options are considered as we are already planning to fully integrate the slash library into dev-master in the near future. If for whatever reason we find those solutions to be insufficient or weighty on the operation of the library in a development setting we will consider option 4 as a possible solution. We will only consider option 5 as a last resort.

I hope this all makes sense to you, but if you disagree with my analysis I would love to hear your feedback and concerns so that I can bring it up in  internal discussions.

shehi:
@valzargaming , I understand where you come from, but this situation will only make things worse, mark my words. My remarks to your comments (thank you for them) would be:

> Use no other dependency at all, maintaining bit conversion methods ourselves within the library using methods such as with base_convert()

We should stick to number-based representation as much as possible, and provide number-based values at least in comments wherever applicable, instead of outright changing them (I mean shift-values such as `1 << x` vs octal-representation nightmare which is very hard to proof-read and debug).

> Use brick/math, which is also MIT licensed, and even the author of the library recommends just using GMP
> Use the GMP extension (discordphp/slash already requires this so there's no issue there)

I absolutely don't understand why `brick/math` needs to be brought in. We can easily stick to one of `php-gmp` or `php-bcmath`. I understand at the very least we have the former already, so question is: what does `brick/math` give us that extension doesn't have? Concrete one or two examples would be great.

> Remove all existing 32-bit support in the library and only support 64-bit, thus breaking at least a few known applications

I don't think this should be last resort, but rather a planned move for a future release. BC complications aren't strange for a major version release. When we have this in mind, all code changes should be made so that all values we change here and there (from 64bit to 32bit), need to have the old 64bit notations at least in the comments, so that we could make switching fast and effectively. I know there are very few maintainers here, but this shouldn't compromise the quality of code at large. As such, I suggest we look forward into the future as well, instead of just trying to survive today :) I hope I could explain this properly, sorry if I couldn't.

> ... cases where an OS may be 32-bit, particularly those that are designed to be cheap and run with a limited amount of memory available ...

IMHO, this shouldn't be relevant to us, as DiscordPHP shouldn't be an embedded software (for IoT or similar). As such, we shouldn't care about such worries that are due to that level of resource limitation.

shehi:
And as we refactor things, let's also write tests guys. Just for things we fix, refactor and similar.

SQKo:
See #661

PR #662, opened 2022-01-01
Fix 32 bit GMP

SQKo:
Fixed wrong return type for void `gmp_setbit`
Update `$is_32_gmp` private, added `Bitwise::is32BitWithGMP()`

Added separate permissions for Stage Instances

### DIFF
--- a/src/Discord/Helpers/Bitwise.php
+++ b/src/Discord/Helpers/Bitwise.php
@@ -16,7 +16,22 @@ namespace Discord\Helpers;
  */
 class Bitwise
 {
-    private static ?bool $is_32_gmp = null;
+    private static bool $is_32_gmp = false;
+
+    /**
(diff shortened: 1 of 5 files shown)


src/Discord/Helpers/Bitwise.php
```diff
@@ -160,4 +162,17 @@ public static function floatCast($value)
 
         return $result;
     }
+
+    public static function is32BitWithGMP(): bool
```
valzargaming, line 166: I'm not crazy about this function being called every single time something would need to use GMP just to determine if the extension is active or not. This should be done only once (preferably on discord __construct()) and the result stored for checking later. This function currently contains branching logic and is considered slow.

valzargaming (changes requested):

valzargaming (approved):

PR #675, opened 2022-01-08
Fix permission errors

### LINKED ISSUE #674: Bot suddenly loses permission to post to channel (without any changes!)
**Environment**
- PHP Version:
    - 8.0
- DiscordPHP Version:
    - dev-master 74fbd0e

**Describe the bug**

I have a very weird problem.

My bot has the posting permission for a specific channel. After some time/restarts the bot suddenly (without ANY changes to the bot or the channel!) loses the ability to post and I get the message:

_There was an error sending the message: You do not have permission to send messages in the specified channel._

Sometimes this could be fixed by simply saving the channel's permissions again in discord (without making any change or making a change to some other permission!)

This was driving me nuts.
After weeks of investigating, I come to the conclusion that it has to be on DiscordPHP side, because DiscordPHP checks the permissions by its own before sending the message.

According to Discord's documentation the permissions are applied in this order:

```
1) Apply denies of @everyone on channel
2) Apply allows of @everyone on channel
3) Sum up all the denies of a member's roles and apply them at once
4) Sum up all the allows of a member's roles and apply them at once
5) Apply denies for a specific member if they exist
6) Apply allows for a specific member if they exist
```
So the order of roles should make no difference, because first all denies are summed up, then all allows.
Does DiscordPHP maybe mess up the order of this permission hierarchy?

To clarify the current situation:
The bot has 2 roles, let's say A and B.

The channel has these permissions for "posting messages":
everyone: deny
A: deny
B: allow

According to the Discord documentation, the bot should have the permission to post a message, because the "allow" from role "B" is applied at the very end.
So maybe DiscordPHP applies the permissions sometimes in the wrong order, e.g. first from role "B" (allow) and then from role "A" (deny)?

I tried to fix this problem and changed the channels permissions to:
everyone: deny
A: neutral
B: allow

This seems to work permanently, but the underlying problem is still there.

SQKo:
This attempts to fix #674 issue with `getPermissions($channel)`, as well correcting the permissions grouping & phpdoc

Also added `ApplicationCommandOptionType` migration in conversion guide

### DIFF
--- a/V7_CONVERSION.md
+++ b/V7_CONVERSION.md
@@ -97,3 +97,4 @@ If you previously linked [DiscordPHP-Slash](https://github.com/discord-php/Disco
 |Send a follow up message|`$interaction->sendFollowUpMessage('text');`|`$interaction->sendFollowUpMessage(MessageBuilder::new()->setContent('text'));`|
 |Update follow up message|`$interaction->updateFollowUpMessage('message_id', 'text');`|`$interaction->updateFollowUpMessage('message_id', MessageBuilder::new()->setContent('text'));`|
 |Delete follow up message|`$interaction->deleteFollowUpMessage('message_id');`|Delete like normal messages<br/>`$message = $channel->messages['message_id'];`<br/>`$message->delete();`|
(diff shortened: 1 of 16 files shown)


SQKo:
Removed unnecessary null coalescing operator when variable is already checked with `isset()` on previous lines

valzargaming (approved):

PR #597, opened 2021-09-04
Implement gateway application command

SQKo:
This is a unified implementation of DiscordPHP with [DiscordPHP-Slash](https://github.com/discord-php/DiscordPHP-Slash) with **Discord v9 API gateway** method, it is merely a copy-paste from the latter library.

The reason I proposed this is due to the future changes from Discord with Application Commands, and your plan in favor of dropping DiscordPHP-Slash to be easily maintained, as well easier for bot coders to integrate. (No hate for the existing discordphp-slash)

~~Since it is a little confusing to decide, I target this for the develop branch instead of master until approved (also to avoid breaking changes with the main branch)~~.

### Main changes
1. No more **RegisterClient**, commands can be now registered through DiscordPHP **application** (as long the bot is authorized with `application.commands` scope) using factory methods.
2. Global commands are now modeled as GlobalCommandRepository `$discord->application->commands`
3. Guild commands are now modeled as GuildCommandRepository `$guild->commands` although I was quite unsure if this pattern is acceptable, it _looked like_ the guild commands repository are not meant for just the bot's commands, i.e. the `application_id` in Guild are meant for the application id of bots created guild (which is rare anyway), I had to workaround the `getRepositoryAttributes()` to override `application_id` with the Bot's application id as it seems another repository does not bind to `:application_id`. However this is the easiest and working approach I could done.
4. Commands are not immediately saved into application or guild repository, they can be freshen/fetch manually when necessary by the coders. It will be automatically cached upon interaction from commands.
4. `registerCommand` is now method of Discord and no longer need to be linked as previously in d.php-slash
5. Filled the part class **Command** (it is empty class in DiscordPHP-Slash), new class **Option** (a copy of Choices with modified structure for [Application Command Option](https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-option-structure), unsure If I'm doing it right)
6. Copied `handleInteraction` and `handleApplicationCommand` codes into `Event::INTERACTION_CREATE` with some modifications (added cache)
7. New event handlers Application Command Create, Application Command Update, Application Command Delete (seems to work only on Global Commands, untested)

### Files/classes migrated from DiscordPHP-Slash to DiscordPHP
- `src/Discord/Enums/ApplicationCommandOptionType.php` -> `src/Discord/Enums/ApplicationCommandOptionType.php` (New: Mentionable and Number type)
- `src/Discord/Parts/Choices.php` -> `src/Discord/Parts/Interactions/Command/Choice.php`
[...]

PR #676, opened 2022-01-08
Add listenCommand auto complete callback

SQKo:
This PR is a followup to #597 especially for autocomplete options, which adds extra callable in `listenCommand()`

Example;
```php
$discord->listenCommand('my_cool_command', function (Interaction $interaction) use ($discord) {
    // do something...
}, function (Interaction $interaction) use ($discord) {
    // check the specific option value here...
    if ($interaction->data->options['optionname']->value) {
        // return suggestions, equivalent to calling $interaction->autoCompleteResult()
        return [
            new Choice($discord, ['name' => 'Choice 1', 'value' => 'one']),
            new Choice($discord, ['name' => 'Choice 2', 'value' => 'two']),
        ];
    }

    // return no result
    return [];
});
```

### DIFF
--- a/src/Discord/Discord.php
+++ b/src/Discord/Discord.php
@@ -1523,12 +1523,13 @@ class Discord
     /**
      * Registeres a command with the client.
      *
-     * @param string|array $name
-     * @param callable     $callback
+     * @param string|array  $name
+     * @param callable      $callback
+     * @param callable|null $autocomplete_callback
      *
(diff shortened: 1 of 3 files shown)


SQKo:
To be noted that At first i wanted it to pass focused option, however there is currently a bug in discord with multiple options having autocomplete.

valzargaming (approved):

PR #677, opened 2022-01-09
Correction for Throwable, phpdoc, add missing attributes

SQKo:
This PR tries to add exception where possible, as well correcting existing throwable and its phpdoc

Based on this https://www.php.net/manual/en/spl.exceptions.php#spl.exceptions.tree the closest exception should be used, for example the command builder had `RangeException` while it should be `LengthException` to adhere the `LogicException` where the code must be fixed at compile time and not `RuntimeException` where the code may have error during run time, due to command builder is being written at compile time.

A `RunTimeException` inside `Promise` of a `Http` request should be also changed to `reject()` instead of `throw`

Also for checking permission if creating request is allowed or not.

Currently have completed with Interactions (Commands and Message Components), will check other classes later.

Also added missing attributes from several Part classes, the phpdoc property attributes is sorted following the API Documentation, and the fillable is sorted by the JSON payload data

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2021 David Cole <david.cole1340@gmail.com> and all contributors
+Copyright (c) 2022 David Cole <david.cole1340@gmail.com> and all contributors
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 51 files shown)


SQKo:
PR is almost done, just need testing after reviews... (Don't merge it yet after approval)

valzargaming (approved):

valzargaming (approved):

SQKo (commented):
I've reviewed and did quick runtime test, and can be merged if there are no more problems.

valzargaming (approved):

PR #678, opened 2022-01-11
Fix repository cleanup on refresh

jocafamaka:
When updating the repository the current items were not being cleaned up correctly (as expected?), leading to a state where previously deleted items still exist in the repository, giving possibility of problematic errors, in all repositories.

It looks like the issue was created in [2016 with the insertion of collections](https://github.com/discord-php/DiscordPHP/commit/4377f9921f3dac0208489bd6488b3979428b1dc6#diff-7170c8fcc06bbd50f347822012ff086a11e63a510672898b30ff440716d1f99eR89).


SQKo:
> It looks like the issue was created in [2016 with the insertion of collections](https://github.com/discord-php/DiscordPHP/commit/4377f9921f3dac0208489bd6488b3979428b1dc6#diff-7170c8fcc06bbd50f347822012ff086a11e63a510672898b30ff440716d1f99eR89).

No, it seems like the problem affected in 2020 where we moved to our own Collection implementation, where previously in 2016 it depended on laravel.

If you look at this commit:

https://github.com/discord-php/DiscordPHP/commit/01a1c0239cab80c557ad1c838549ccaecf267f6a

you will see that your PR makes sense and old code repistory doing fill([]) was correct


However, we need to find more codes that might affect this before just merging away with that conclusion...

jocafamaka:
For reference, this is the commit that changed the implementation of the fill function.

https://github.com/discord-php/DiscordPHP/commit/75af83942348013232b4b9073d27953ea6dff68b#diff-a5ffd948944ac6f75406fdeefa56f3e0d81902dbc07d7001507eba272c70281dR92

SQKo:
> However, we need to find more codes that might affect this before just merging away with that conclusion...

Confirmed the change is safe to commit

SQKo (approved):

PR #700, opened 2022-01-29
Create cmdgen-example.php

givinghawk:
All this really does is send a codeblock with a command, just to save writing the same lines a bunch of times


SQKo:
Have you ever checked the [DiscordCommandClient](https://github.com/discord-php/DiscordPHP/blob/master/src/Discord/DiscordCommandClient.php) class? It help you to create commands easily, although not in generating code in your context. But I think example using [DiscordCommandClient](https://github.com/discord-php/DiscordPHP/wiki/Command-Client) would suit.

givinghawk:
I will try to put that in here

PR #711, opened 2022-02-04
fix potential object event bug, and variable naming

SQKo:
The events affected were my works of my previous PR, I noticed potential bugs when trying to document them, but there is no parameter change so there is no breaking change. Also renamed the local variable names to studly.


valzargaming (dismissed):

SQKo:
The GUILD_SCHEDULED_EVENT_USER_ADD and GUILD_SCHEDULED_EVENT_USER_REMOVE somehow does not work, as it is locked to 3 arguments, so it's breaking change (it was never working anyway). Now only the raw event $data is returned in the first argument of both events plus the Discord object.

valzargaming (approved):

PR #694, opened 2022-01-25
Add user caching from event

SQKo:
#518 

Setter is necessary to apply when explicitly send from the event.

- ~~CHANNEL_CREATE~~
- ~~CHANNEL_DELETE~~
- ~~CHANNEL_PINS_UPDATE~~
- ~~CHANNEL_UPDATE~~
- [x] GUILD_BAN_ADD
- [x] GUILD_BAN_REMOVE
- _**GUILD_CREATE**_
- ~~GUILD_DELETE~~
- [x] GUILD_EMOJIS_UPDATE
- ~~GUILD_INTEGRATIONS_UPDATE~~
- [x] **GUILD_MEMBER_ADD**
- [x] GUILD_MEMBER_UPDATE
- [x] GUILD_MEMBER_REMOVE
- ~~GUILD_ROLE_CREATE~~
- ~~GUILD_ROLE_UPDATE~~
- ~~GUILD_ROLE_DELETE~~
- [x] GUILD_SCHEDULED_EVENT_CREATE
- [x] GUILD_SCHEDULED_EVENT_DELETE
- [x] GUILD_SCHEDULED_EVENT_UPDATE
- ~~GUILD_SCHEDULED_EVENT_USER_ADD~~
- ~~GUILD_SCHEDULED_EVENT_USER_REMOVE~~
- [x] GUILD_STICKERS_UPDATE
- ~~GUILD_UPDATE~~
- [x] INTEGRATION_CREATE
- ~~INTEGRATION_DELETE~~
- [x] INTEGRATION_UPDATE
- [x] **_INTERACTION_CREATE_**
- [x] _INVITE_CREATE_
- ~~INVITE_DELETE~~
- [ ] _MESSAGE_CREATE_?
- ~~MESSAGE_DELETE~~
- ~~MESSAGE_DELETE_BULK~~
- [x] _MESSAGE_REACTION_ADD_
- ~~MESSAGE_REACTION_REMOVE~~
- ~~MESSAGE_REACTION_REMOVE_ALL~~
- ~~MESSAGE_REACTION_REMOVE_EMOJI~~
- [ ] _MESSAGE_UPDATE_?
- ~~_**PRESENCE_UPDATE**_~~
- ~~STAGE_INSTANCE_CREATE~~
- ~~STAGE_INSTANCE_DELETE~~
- ~~STAGE_INSTANCE_UPDATE~~
- [ ] ?_THREAD_CREATE_
- ~~THREAD_DELETE~~
[...]

PR #715, opened 2022-02-06
Fix most of event cache data inconsistency, variables naming, and add docs link

### LINKED ISSUE #714: joined_at is always null when using loadallmembers
**Environment**
- PHP Version:
    - 8.1.2
- DiscordPHP Version:
    - dev-master

**Describe the bug**
Everytime when you use `$member->joined_at` it is null

**To Reproduce**
```php
$discord->on(Event::GUILD_MEMBER_REMOVE, function (Member $member, Discord $discord) {
    $logchannel = $discord->getChannel("<channel id>");
    $logchannel->sendMessage("joined_at: {$member->joined_at}");
});
```

**Expected behavior**
Get the join date of the member

SQKo:
Follow up to #711 but for old & existing events

Fixes #714 (only if member is cached)

I will also check for other events

### DIFF
--- a/src/Discord/WebSockets/Events/GuildMemberRemove.php
+++ b/src/Discord/WebSockets/Events/GuildMemberRemove.php
@@ -15,6 +15,9 @@ use Discord\Parts\User\Member;
 use Discord\WebSockets\Event;
 use Discord\Helpers\Deferred;
 
+/**
+ * @see https://discord.com/developers/docs/topics/gateway#guild-member-remove
+ */
 class GuildMemberRemove extends Event
 {
     /**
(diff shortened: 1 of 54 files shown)


src/Discord/WebSockets/Events/GuildMemberRemove.php
```diff
@@ -22,13 +22,17 @@ class GuildMemberRemove extends Event
...
+        if ($guild = $this->discord->guilds->get('id', $data->guild_id)) {
+            $member = $guild->members->pull($data->user->id);
             --$guild->member_count;
+        }
 
-            $this->discord->guilds->push($guild);
+        if (! $member) {
+            /** @var Member */
+            $member = $this->factory->create(Member::class, $data, true);
+            $member->guild_id = $data->guild_id;
```
SQKo, line 35: Due to GUILD_MEMBER_REMOVE actually only returns an `User`, instead of `Member`, this should be replaced in next major version update.

The question here is whether to attach `guild_id` or not since the member is already removed.

SQKo:
Also follow up to #694 for `MESSAGE_CREATE` event only

valzargaming (approved):

valzargaming (approved):

SQKo:
There was a little mistake with GuildBanRemove, everything else has been tested, so it can be merged right away.

key2peace (approved):

PR #728, opened 2022-02-16
Bump version, deprecate invite stage instance, thread updateable

SQKo:
- Bump version to 7.0.2
- Deprecate stage instance attribute in invite https://github.com/discord/discord-api-docs/pull/4479
- Fix phpdoc `Member::moveMember()` parameter `$channel` nullable
- Add updateable attributes for `Thread`, the following functions are separate to be used by Bots without `MANAGE_THREAD` permissions:
  - Add `Thread::rename()`
  - Add `Thread::setAutoArchiveDuration()`
  - Add audit log `$reason` parameters for `Thread::archive()` and `Thread::unarchive()`

Tested.

EDIT: also added Attachment stuff to `MessageBuilder`, this seems required to keep attachment in editing messages or not to append them when creating messages. (The behavior is required in Discord API v10)

### DIFF
--- a/src/Discord/Builders/MessageBuilder.php
+++ b/src/Discord/Builders/MessageBuilder.php
@@ -17,6 +17,7 @@ use Discord\Builders\Components\SelectMenu;
 use Discord\Exceptions\FileNotFoundException;
 use Discord\Helpers\Multipart;
 use Discord\Http\Exceptions\RequestFailedException;
+use Discord\Parts\Channel\Attachment;
 use Discord\Parts\Channel\Message;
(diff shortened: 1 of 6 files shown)


SQKo:
Also added `CLOSE_NORMAL` code 1000 with `$discord->close()`, but since it's technically impossible to send the code, you have to set force close to false and close the loop after the opcode has been send.

This is how to exit bot gracefully and make it offline:
```php
$discord->close(false);
$loop = $discord->getLoop();
$loop->futureTick([$loop, 'stop']);
```

PR #738, opened 2022-02-24
Fixed null-pointer exception case.

shehi:
Fixed null-pointer exception case. Otherwise, a method like `addChoice()` in the same file which checks against `count($this-choices)` will get `null` value as an input and cause exception.

image

image

### DIFF
--- a/src/Discord/Parts/Interactions/Command/Option.php
+++ b/src/Discord/Parts/Interactions/Command/Option.php
@@ -65,12 +65,12 @@ class Option extends Part
     /**
      * Gets the choices attribute.
      *
-     * @return Collection|Choice[]|null A collection of choices.
+     * @return Collection|Choice[] A collection of choices.
      */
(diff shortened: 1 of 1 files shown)


SQKo:
It was left to be nullable since other type of commands can't have choices. There is no way from the class code to check if the option from a slash command or not.
So you have to manually to check it from your end using `isset($this->choices)`, `! empty($this->choices)`, `count($this->choices ?? [])` or simple `if ($option->choices)`

shehi:
Totally understandable, yet this makes a smelly code. We gotta find solution to this...

Then at the very least, I can fix LINE-220 where `$this-choices` is being queried.

SQKo:
> 

You are right about line 220, but back again to the reason I mentioned, there is no way to check the option being added into a slash command, so addChoice() is supposed to be illegal/throw error in non slash commands.

shehi:
Close #739 if that still doesn't work :) You definitely know the codebase better.

PR #739, opened 2022-02-24
Fixed null pointer exception case.

shehi:
Potential null pointer exception case fix. Hope this is acceptable.

### DIFF
--- a/src/Discord/Parts/Interactions/Command/Option.php
+++ b/src/Discord/Parts/Interactions/Command/Option.php
@@ -25,7 +25,7 @@ use function Discord\poly_strlen;
  * @property string                   $name          Name of the option.
  * @property string                   $description   1-100 character description.
  * @property bool                     $required      If the parameter is required or optional--default false.
- * @property Collection|Choice[]|null $choices       Choices for STRING, INTEGER, and NUMBER types for the user to pick from, max 25.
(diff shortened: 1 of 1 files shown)


shehi:
I don't imagine how this fix could ruin anything: `count()` requires iterable, so if somehow we have null value here, it needs to coalesce to an empty array. If we need to limit the usage of `addChoice()` method in non-slash-command contexts, then we need to find a different approach.

shehi:
Yea, the request will be rejected. Anyone following the Discord docs wouldn't make that mistake anyway. And that rejection is reflected in Bot logs, right?

SQKo:
> Yea, the request will be rejected. Anyone following the Discord docs wouldn't make that mistake anyway. And that rejection is reflected in Bot logs, right?

It should, depends on how you handle the rejected promises, due to possible ghost exception issue that will get fixed until ReactPHP/3.0, however this is only for builder functions, so developer should be aware they are not adding choice to non slash commands.

The same actually also applies to setType 11 (attachment) it cannot be used elsewhere other than main option of slash command.

Passing/filling empty array in this class is not the solution because this class is mainly used for Parts, that is the data received from interactions, the builders functions were there only for convenience. We will probably rework on the builder class instead of having the functions here.

SQKo (approved):

PR #759, opened 2022-03-09
Threads inherit permissions from the channel in which they were created

valzargaming:
https://support.discord.com/hc/en-us/articles/4404809613847-Threads-Moderation-FAQ#:~:text=Generally%2C%20threads%20inherit%20permissions%20from,can%20send%20messages%20in%20Threads.

### DIFF
--- a/src/Discord/Parts/User/Member.php
+++ b/src/Discord/Parts/User/Member.php
@@ -311,7 +311,7 @@ class Member extends Part
         if ($channel instanceof thread) {
             $channel = $this->guild->channels->get('id', $channel->parent_id);
         } elseif (!($channel instanceof channel)) {
-            return reject(new \RuntimeException('Must be an instace of Channel or Thread.'));
+           return $this->factory->part(RolePermission::class, ['bitwise' => 0]);
         }
         // Get @everyone role guild permission
         $bitwise = $this->guild->roles->get('id', $this->guild_id)->permissions->bitwise;
--- a/src/Discord/Parts/User/Member.php
+++ b/src/Discord/Parts/User/Member.php
@@ -302,16 +302,16 @@ class Member extends Part
      *
      * @see https://discord.com/developers/docs/topics/permissions
      *
-     * @param Channel|null $channel
+     * @param Channel|Thread|null $channel
      *
      * @return RolePermission
      */
     public function getPermissions($channel = null): RolePermission
     {
-        if ($channel instanceof thread) {
+        if ($channel instanceof Thread) {
             $channel = $this->guild->channels->get('id', $channel->parent_id);
-        } elseif (!($channel instanceof channel)) {
-           return $this->factory->part(RolePermission::class, ['bitwise' => 0]);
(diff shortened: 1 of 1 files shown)


valzargaming:
Tested and confirmed working, can be pulled right away

valzargaming:
There are two additional cases that have not yet been accounted for:

If they have Send Messages in Threads permission enabled, then they can send messages in Threads.
If they have Create Public Threads or Create Private Threads permission enabled, they can create a thread from an existing message or scratch.

src/Discord/Parts/User/Member.php
```diff
@@ -305,8 +306,13 @@ public function sendMessage($message, bool $tts = false, $embed = null, $allowed
      *
      * @return RolePermission
      */
-    public function getPermissions(?Channel $channel = null): RolePermission
+    public function getPermissions($channel = null): RolePermission
     {
+        if ($channel instanceof thread) {
```
davidcole1340, line 311:
```suggestion
        if ($channel instanceof Thread) {
```

src/Discord/Parts/User/Member.php
```diff
@@ -305,8 +306,13 @@ public function sendMessage($message, bool $tts = false, $embed = null, $allowed
...
+            $channel = $this->guild->channels->get('id', $channel->parent_id);
+        } elseif (!($channel instanceof channel)) {
+           return $this->factory->part(RolePermission::class, ['bitwise' => 0]);
+        }
```
davidcole1340, line 315: This will prevent `$channel` being passed as `null`. Regardless, I think it should be changed to throw an exception in the case `$channel` is not `Channel`, `Thread` or `null`.

davidcole1340 (changes requested):
Can we also update the docblock typehint for `$channel` as well? `@param Channel|Thread|null $channel`.

src/Discord/Parts/User/Member.php
```diff
@@ -301,12 +302,17 @@ public function sendMessage($message, bool $tts = false, $embed = null, $allowed
...
      * @return RolePermission
      */
-    public function getPermissions(?Channel $channel = null): RolePermission
+    public function getPermissions($channel = null): RolePermission
     {
+        if ($channel instanceof Thread) {
+            $channel = $this->guild->channels->get('id', $channel->parent_id);
+        } elseif ($channel && !($channel instanceof channel)) {
+            return $this->factory->part(RolePermission::class, ['bitwise' => 0]);
+        }
```
davidcole1340, line 315:
```suggestion
        } elseif ($channel && !($channel instanceof Channel)) {
            throw new \InvalidArgumentException('$channel must be an instance of Channel, Thread or null.');
        }
```

valzargaming:
Current code only fixes exception thrown when attempting to use getPermissions within a thread, which may be the default behavior for some bots currently utilizing the `$discord->on('message'` intent.

SQKo (approved):

davidcole1340 (approved):

PR #761, opened 2022-03-13
Need a hasRole() function -> Member part

osemk:
Need a $member->hasRole("id") function to check member roles one by one.


key2peace:
please fork this project, then create a branch on your fork, apply the changes you want in that branch and then submit pr from there

SQKo:
there is already `$member->roles->has(...)`

PR #796, opened 2022-04-30
Add a way to check status for Interaction::responded from get() functions

### LINKED ISSUE #794: Interaction object cannot be fully restored
**Environment**
- PHP Version:
    - 8.0
- DiscordPHP Version:
    - latest dev

**Describe the bug**
With the old slash library it was possible to store an Interaction object (Discord\Slash\Parts\Interaction) in a json, restore it later (even after a bot restart) and change the original message with "updateInitialResponse":
```
$data=json_encode($interaction->jsonSerialize());
[...]
$interaction=new Interaction(json_decode($data),$discord);
$interaction->updateInitialResponse(...);
```
After I removed the slash library and switched to the new native commands, the object can still be restored:
```
$data=json_encode($interaction->getRawAttributes());
[...]
$interaction=new Interaction($discord,json_decode($data,TRUE));
$interaction->updateOriginalResponse(...);
```
But with one big problem: The native Interaction class has the internal variable "responded". This variable isn't set through the attributes and there is no way to set this variable from outside (because it's protected and has no setter).
But to use "updateOriginalResponse" with the restored object the variable "responded" must be "true", otherwise the posting won't update.

For a quick fix I added this to Discord/Parts/Interactions/Interaction.php:
```
    public function setResponded()
    {
        $this->responded=true;
    }
```
When I use $interaction->setResponded() after the restore of the Interaction object, now everything works as expected.
But an "official" way to do this would be nice. 
Thanks for your support.

SQKo:
This changes the behavior of `getOriginalMessage()` & `getFollowUpMessage()` to be used when interaction has been not responded yet, however that will just return error from API if that is the case.

Fixes #794, will wait for positive test and feedback first before merge.

### DIFF
--- a/src/Discord/Parts/Interactions/Interaction.php
+++ b/src/Discord/Parts/Interactions/Interaction.php
@@ -272,12 +272,13 @@ class Interaction extends Part
      */
     public function getOriginalResponse(): ExtendedPromiseInterface
     {
-        if (! $this->responded) {
-            return reject(new \RuntimeException('Interaction has not been responded to.'));
+        if (! $this->created) {
+            return reject(new \RuntimeException('Interaction has not been created yet.'));
         }
 
         return $this->http->get(Endpoint::bind(Endpoint::ORIGINAL_INTERACTION_RESPONSE, $this->application_id, $this->token))
             ->then(function ($response) {
+                $this->responded = true;
                 return $this->factory->create(Message::class, $response, true);
             });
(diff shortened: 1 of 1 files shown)


Swift42 (approved):
For me it's a good and working solution.
(I tested the changes of "getOriginalResponse")

valzargaming (approved):

PR #798, opened 2022-05-04
fix command options

HolySmoke86:
correctly cast comand options from OptionRepository to array when
passing to RegisteredCommand::execute()


SQKo:
Sorry, I don't know what is the "fix" here, can you show what's wrong? The collection class implements ArrayAccess so it should not need to be converted to array.

HolySmoke86:
Well, there's two issues at play here:
The first one being the expression `$option['options'] ?? []` always evaluating to `[]`, which prevents options from being passed to commands (and subsequently subcommands from being executed, which was the issue I had).
The second one is the function signature type hints the options parameter as `array` which only accepts actual arrays and throws when being passed an `OptionRepository`.

A concrete example: I have a subcommand registered with
```
$discord->listenCommand(['aosr', 'preset'], function(Interaction $interaction) use ($discord) {
   // ...
});
```
the interaction handler of which will not be called without above patch.
Applying only the coalesce operator portions results in a `TypeError`, adding the array cast successfully executes the subcommand's callback.

SQKo:
>  The second one is the function signature type hints the options parameter as array which only accepts actual arrays and throws when being passed an OptionRepository.

Can you remove the array type there and try if it works?

> The first one being the expression $option['options'] ?? [] always evaluating to []

No it's because the sub command options is treated differently in the Part side.
I actually still havent figured it out since this was ported from DiscordPHP-Slash, but I've been guessing theres something that need to deal with `$option['type']` on `$checkCommand` for sub command and sub command groups. The codes were only for normal commands options

Also, Can you post the command register codes so i can reproduce it?

HolySmoke86:
If I replace the function signature to
```
public function execute(array|\ArrayAccess $options, Interaction $interaction): bool
```
I can do aways with the `->toArray()` call.
Can't get it to work with the `??` operator though. Looks like it's supposed to call `Collection::offsetExists` for the implied `isset` check, but it always behaves like that returned false, so always evaluates to `[]`.

The full command registration and implementation is here: https://git.localhorst.tv/?p=alttp.git;a=blob;f=app/DiscordAppCommands/AosrPresetCommand.php
It's just a basic subcommand like the example from the discord developer portal.

SQKo:
> I can do away with the `->toArray()` call.
changing this with ->toArray() would break the existing user implementations that deals $options repository with collections functions, especially to non sub command options.
The correct way is to check first if the option command is a sub command or not.

> Can't get it to work with the `??` operator though. Looks like it's supposed to call `Collection::offsetExists` for the implied `isset` check, but it always behaves like that returned false, so always evaluates to `[]`.
This seems to be issue with array_key_exists that has been removed on php 8.0
https://www.php.net/manual/en/function.array-key-exists.php (see notes)

Likely the addSubcommand will be fixed or removed, depends on the difficulty since Discord slash commands applies special rules that command options can be nested in some levels only and will not allow the same level to not have sub command options. You can assume the addSubcommand was there for legacy implemenation of DiscordPHP-Slash, which i don't remember how it was, but i believe it is not for the current implementation of the API.

So for now if you want to handle subcommand, don't pass array on listenCommand (i.e. `listenCommand(['aosr', 'preset'])`), instead just use listenCommand on the base command `listenCommand(aosr)` and handle the $interaction->data->options manually to determine the sub command options.

> The full command registration and implementation is here: https://git.localhorst.tv/?p=alttp.git;a=blob;f=app/DiscordAppCommands/AosrPresetCommand.php It's just a basic subcommand like the example from the discord developer portal.

Thanks for this I'll test out later when i have time

SQKo (approved):

SQKo:
Note to myself: Autocomplete listeners will only work on base command.

PR #818, opened 2022-05-28
Gateway Ready no longer sends private channels

key2peace:
As per <https://discord.com/developers/docs/topics/gateway#ready>, the ready event no longer transmits private channels in it's data, thus making this code obsolete. The option pmChannels however is kept in order for the user to decide wether to store newly created private channels or not.


SQKo (approved):
The pmChannels is not used elsewhere, can be removed in next major version

PR #824, opened 2022-06-15
Few fixes in threads

jocafamaka:
Allow create threads with null `auto_archive_duration` for nom auto-archiving threads. (https://github.com/discord/discord-api-docs/commit/fb317e48cd868a5813c8e47c4ae09060d7c56bcf)

Improve channel type validation for [threads from message](https://discord.com/developers/docs/resources/channel#start-thread-from-message).


jocafamaka:
It seems that there is still no possibility in the lib to delete threads or edit some properties, I will work to include it in this PR.

SQKo (approved):

Issue #493, opened 2021-02-10
Private channels not stored correctly


rederrik:
**Environment**
- PHP Version:
    - 7.4.5
- DiscordPHP Version:
    - dace512

**Describe the bug**
Title, I can't seem to get private channels by their ID using `Discord::getChannel`. Additionally, breakpoints set to every instance of `private_channels->push` never fire.
I know I can get private channels by other means, but I need this method for generic delayed events.

**To Reproduce**
```php
$discord->registerCommand(
    'pmtest',
    function (Message $message) use ($discord) {
        $chId = $message->channel_id;
        if ($channel = $discord->getChannel($chId)) {
            $channel->sendMessage('direct');
        } else {
            $message->reply('cant get private channel');
        }
    }
);
```
Then call this command in private channel with the bot - you should receive the 'cant get private channel' reply since `private_channels` repo is empty. `pmChannels` option value doesn't matter.

**Expected behavior**
Get 'direct' message  after firing this command in a private channel with the bot.

davidcole1340:
Discord doesn't provide any way to retrieve existing private message channels, so they have to be "created" (actually returns the same channel every time) after every restart of the bot. When you call `$user->getPrivateChannel()` the channel will be cached so that it doesn't have to be "created" again. Is this what you were meaning?

rederrik:
I mean, when the bots receives a command as a private message the channel is already being "created", isn't it? In my example, the `reply` to that channel works, but getting that channel from discord client doesn't.

davidcole1340:
I just added a change on the `develop` branch, private channels will now be created if they don't exist. That's basically all we can do, we aren't able to get a list of private channel IDs from Discord, we have to wait until a user messages the bot.

rederrik:
I'm confused with that change honestly, as my example still fails to get the channel from the client.

I tried this instead and it seem to work: 
```php
if(($channel = $message->channel) && ($channel->type === Channel::TYPE_DM)) {
    $this->discord->private_channels->push($channel);
}
```

Then I thought, "why do this in `MessageCreate` event?". So I reverted that change, went to `ChannelCreate` and changed
```php
if (isset($data->attributes['recipients'])) {
```
to
```php
if ($channel->type === Channel::TYPE_DM) {
```
And it worked too!

But why is `ChannelCreate` handler being run on every message?

davidcole1340:
[...]

PR #825, opened 2022-06-16
make sure to populate private_channels when receiving PMs

putnam:
When receiving a PM, the $discord->private_channels array sometimes does not have the DM channel added. If the channel is later needed (e.g., in an event handler), getChannel will fail because the DM channel was not inserted.

This covers the case of "a channel already exists", whereas "the channel does not exist yet" was already covered by the stanza above.

See #493 which was partially fixed last year.

### DIFF
--- a/src/Discord/Repository/PrivateChannelRepository.php
+++ b/src/Discord/Repository/PrivateChannelRepository.php
@@ -28,7 +28,9 @@ class PrivateChannelRepository extends AbstractRepository
     /**
      * @inheritdoc
      */
-    protected $endpoints = [];
+    protected $endpoints = [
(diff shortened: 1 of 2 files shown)


SQKo:
Thank you for raising back the issue. After taking a quick look, it seems that your PR kinda misleading. I think this line is actually the cause, because this is legacy code when private channels was a thing in Discord API:
```php
        // assume it is a private channel
        if ($messagePart->channel === null) {
```

Anyway I'll take a look and test it later to see the correct solution and update this PR

SQKo (approved):
Added a way to fetch DM by ID `$discord->private_channels->fetch('id', ...);`

Also Fixes #493 again

PR #846, opened 2022-07-02
Allow Symfony 6 and Monolog 3

kylekatarnls:
Increase compatibility to allow installation inside projects with more modern stacks.

OptionsResolver should not be an issue as it's just stricter types:
https://github.com/symfony/options-resolver/compare/v5.4.3...v6.0.0

Monolog is at a slightly higher risk (introduction of the enum `Level` and `format()` but they're not used in this package so it should be compatible:
https://github.com/Seldaek/monolog/compare/2.7.0...3.0.0


composer.json
```diff
@@ -13,9 +13,9 @@
         "nesbot/carbon": "^2.38",
         "ratchet/pawl": "^0.4.1",
         "react/datagram": "1.5.*",
-        "symfony/options-resolver": "^5.1.3",
+        "symfony/options-resolver": "^5.1.3 || ^6.0",
         "trafficcophp/bytebuffer": "^0.3",
-        "monolog/monolog": "^2.1",
+        "monolog/monolog": "^2.1 || ^3.1",
```
SQKo, line 18: Any particular reason requiring 3.1 instead of 3.0?

SQKo:
Everything else looks good to me, but I've not been able to test yet with machine on php 8.0 & php 8.1

composer.json
kylekatarnls:
```suggestion
        "monolog/monolog": "^2.1 || ^3.0",
```

kylekatarnls: Nope, it should be no issue.

SQKo (approved):

PR #847, opened 2022-07-02
Use arrow functions for single-step transformations

kylekatarnls:


SQKo (changes requested):
Please leave the codes dealing with `Collection` as is, beside of considering how `array_map()` would impact performance, we also wanted to add more codes in those block to deal with cache in future (so it can't be those one liner).

kylekatarnls:
> array_map() would impact performance

It's faster actually than using a foreach loop and push each item.

I also recommend when having a lot of code in the callback to create a dedicated method:

Example:
```php
    public function getPinnedMessages(): ExtendedPromiseInterface
    {
        return $this->http->get(Endpoint::bind(Endpoint::CHANNEL_PINS, $this->id))
            ->then(fn ($responses) => $this-getMessagesCollection($responses));
    }

    private function getMessagesCollection(array $responses): Collection
    {
        return  new Collection(array_map(
            fn ($response) => $this->messages->get('id', $response->id)
                ?: $this->factory->create(Message::class, $response, true),
            $responses,
        )));
    }
```

And in this case you can make `getMessagesCollection` evolve, same logic can be applied to `$this->messages->get('id', $response->id) ?: $this->factory->create(Message::class, $response, true)` => have a dedicated method.

But feel free to close if you prefer bigger methods over more methods, it's just a preference matter.

SQKo:
Closing this because we are going to rewrite the Repository class for the upcoming cache interface

PR #873, opened 2022-08-04
Fix Events

SQKo:
The `GUILD_DELETE` event does not send a `Guild` object in `$data` even if it was just unavailable (reference: https://github.com/discord-payloads/discord-payloads/blob/master/events/guild/guild_delete-unavailable.json) so the Part should be left null for uncached Guild (this is already noted in our docs)

Reference: https://discord.com/developers/docs/topics/gateway#guild-delete

This partially reverts https://github.com/discord-php/DiscordPHP/commit/59283c443bc078bd088ac3f59aa5899c17694fb2


SQKo:
Also fix behavior with `GUILD_BAN_REMOVE` event (to prioritize cached bans reference)

key2peace (approved):

Issue #806, opened 2022-05-23
TODO: Breaking Changes in v10


SQKo:
Since the deprecation of bulk endpoints for the BOT in the Permissions v2, the application command permissions "overwrite" structure is no longer relevant to be a repository, also the naming should be just "command Permissions", with "Permission" parts in it.

At the moment the repository only contains a single Permission part

- [x] Rename Remove "Overwrite" to "CommandPermissions"
- [x] Change type from Repository to Part
- [x] Remove disabled endpoints that only work with bearer oauth2 token

I think this can be more to swapping the Repository->Part with Part->Repository, I'm not sure yet we'll see later

- [x] It's also seem that more stuff need 64 bits calculation, so GMP extension will be required in v8.0 for 32 bits PHP and will exit instead of warning the bot developer. #960 

Since this is a breaking change, I'm noting this for v8 milestone.

EDIT:
- [x] Also remove `pmChannels` options, see ~~#818~~ #887

SQKo:
The guild feature `SEVEN_DAY_THREAD_ARCHIVE`  and `THREE_DAY_THREAD_ARCHIVE` should be also removed in next major version, deprecate in next minor version. (these restriction is no longer present)

SQKo:
The `$message->channel_mentions` behavior should be changed that is actually purposed for Channels object as seen in followed News Channel, not actual #channel mention in the message.

SQKo:
Fix visibility inconsistency with non standard attributes such as `Message::getLinkAttribute()` as the rest of `get<>Attribute` are `protected`, not `public` unless having override parameters (e.g. `getAvatarAttribute()`).

SQKo:
`Reaction::getUsers()` no longer have 'before' options, providing this to the API has no effect. Thus the option should be removed.

See https://discord.com/developers/docs/resources/channel#get-reactions

SQKo:
The guild feature `COMMERCE` has been removed along with store channels

https://github.com/discord/discord-api-docs/pull/5180

SQKo:
Stage Instance Repository should be part of Channel, not Guild, even though the events filled from guild, this is same how Discord fills threads from Guild Create event.

SQKo:
We also agreed to remove `$scriptData` from `Part`, Not sure if there are many people using it, please comment here if you do.

This is due to the change to caching system and `Part::scriptData` will be likely Garbage Collected. The solution is to reference the Part object instead.
```php
$guild->scriptData = $myData;
[...]

PR #886, opened 2022-08-11
Reaction::getUsers() no longer have 'before' options

key2peace:
#806


SQKo (approved):
Don't merge yet before v10 pre-release

PR #887, opened 2022-08-11
Remove obsolete pmChannels

key2peace:


SQKo (approved):
Don't merge yet before v10 pre-release

PR #914, opened 2022-09-08
attempt to fix prevent chunking on ready server outage

### LINKED ISSUE #620: Unavailable servers prevent setupChunking()
https://github.com/discord-php/DiscordPHP/blob/cf79ca6ada74102b50ade8f84d99db028d585505/src/Discord/Discord.php#L459

setupChunking never gets called if something goes wrong and a server that is expected to load never loads. Removing this if statement or calling setupChunking() after a set amount of time has passed resolves the issue and allows start-up functions such as with loadAllMembers to process correctly. This issue occurs if an empty guild is sent by Discord but neither a guild create or guild delete event is called for that guild by Discord.

https://discord.com/developers/docs/topics/gateway#guild-availability
https://discord.com/developers/docs/topics/gateway#guild-create

# Before
image

# After
image

SQKo:
Attempt to fix #620 but cannot be confirmed as we do not have server outage at the moment.

Feel free to use the `dev-patch-620` branch or pull the patch code if you face the issue like in #620

Reference: https://discord.com/developers/docs/resources/guild#unavailable-guild-object

### DIFF
--- a/src/Discord/Discord.php
+++ b/src/Discord/Discord.php
@@ -516,6 +516,17 @@ class Discord
         };
 
         $this->on(Event::GUILD_CREATE, $function);
+
+        $function2 = function ($guild) use (&$function2, &$unavailable) {
+            if ($guild->unavailable) {
+                $this->logger->debug('guild unavailable', ['guild' => $guild->id, 'unavailable' => count($unavailable)]);
+                unset($unavailable[$guild->id]);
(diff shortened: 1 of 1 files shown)


key2peace:
wouldn't it be easier to map GUILD_DELETE to $function and remove both listeners when unavailable reached 0 ?

PR #920, opened 2022-09-10
Fix command builders and Promise error handling

SQKo:
Cherry picked from v7 branch ff372aade73141e446f67f86226aef40873c55a3

This add checks for properties that are not set especially when used in trait, which was throwing PHP notice `Indirect modification of overloaded property Discord\Parts\Interactions\Command\Command::$options has no effect` when used in array read + append.
For the Part class, it is set from attributes array instead.

Thanks to mrWorst#0856 for reporting on Discord!

### DIFF
--- a/src/Discord/Builders/CommandAttributes.php
+++ b/src/Discord/Builders/CommandAttributes.php
@@ -75,7 +75,7 @@ trait CommandAttributes
             throw new \LengthException('Command name can be only up to 32 characters long.');
         }
 
-        if (isset($this->type) && Command::CHAT_INPUT == Command::CHAT_INPUT && preg_match('/^[-_\p{L}\p{N}\p{Devanagari}\p{Thai}]{1,32}$/u', $name) === 0) {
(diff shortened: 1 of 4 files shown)


src/Discord/Builders/CommandAttributes.php
```diff
@@ -75,7 +75,7 @@ public function setName(string $name): self
             throw new \LengthException('Command name can be only up to 32 characters long.');
         }
 
-        if ($this->type == Command::CHAT_INPUT && preg_match('/^[-_\p{L}\p{N}\p{Devanagari}\p{Thai}]{1,32}$/u', $name) === 0) {
+        if (isset($this->type) && Command::CHAT_INPUT == Command::CHAT_INPUT && preg_match('/^[-_\p{L}\p{N}\p{Devanagari}\p{Thai}]{1,32}$/u', $name) === 0) {
```
key2peace, line 78: shouldn't that be ``&& $this->type == Command::CHAT_INPUT &&``

key2peace (changes requested):

src/Discord/Builders/CommandAttributes.php
SQKo: Thanks, it was obvious bad auto merge

key2peace (approved):

PR #921, opened 2022-09-10
Update examples to use MESSAGE_CONTENT

jocafamaka:
Make examples great again!

Fix examples to use MESSAGE_CONTENT intent.

### DIFF
--- a/examples/browser.php
+++ b/examples/browser.php
@@ -11,10 +11,11 @@
  * php examples/browser.php
  */
 
-include __DIR__.'/../vendor/autoload.php';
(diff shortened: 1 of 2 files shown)


SQKo:
Already included in #916 but if you wish you can change the target to v7 branch, Although API V9 will just work fine without the Intent set from code

jocafamaka:
I think just changing the target would break things, because I did it in master...

On the other hand, I think it's nice to have a heads-up for new users.

Discord has enabled MESSAGE_CONTENT by default for unverified apps in the enforcement deadline, but I don't know if it's enable by default for new bots...

But leme kill this for now.

PR #937, opened 2022-09-29
feat: Add tests to collections

Exanlv:
~~Adds `->each` method to `Collection` as an alternative to `foreach ($collection as $item)`.~~ removed

Also adds tests for `Collection` as these were lacking


src/Discord/Helpers/Collection.php
```diff
@@ -262,7 +262,7 @@ public function isset($offset): bool
     }
 
     /**
-     * Checks if the array has an object.
+     * Checks if the array has multiple offsets.
```
Exanlv, line 265: Old description wasn't accurate

Exanlv:
I don't find the codacy report to make much sense but I'll see what you guys have to say I suppose

SQKo:
What is your use case with DiscordPHP?

If it is to deal with Repository, the behavior of Collection is deprecated due to limitation in asynchronous code.

Exanlv:
Exporting all messages within a channel is my usecase :^)

I prefer this syntax with Laravel's collection and felt it was missing here

SQKo:
Then this cannot be done since the `dev-master` has major changes with the cache, using a possibly asynchronous flow with Message Repository, which mean each iteration cannot be blocking otherwise you just wont get any next Message in the iteration

SQKo (approved):

PR #953, opened 2022-10-21
Remove rule from PHPCS, exclude examples

Exanlv:
Stop `composer run-script cs` from changing PHPDoc comments to code comments.

Exclude `examples` dir from phpcs fixer


SQKo (approved):
thank

PR #954, opened 2022-10-22
Feature: more function tests

Exanlv:
More tests woo

~~also updated cs cuz my local repo wasnt updated when I PR'd those other changes~~

### DIFF
--- a/tests/FunctionsTest.php
+++ b/tests/FunctionsTest.php
@@ -183,10 +183,12 @@ final class FunctionsTest extends TestCase
 
     public function normalizePartIdProvider(): array
     {
+        $mockDiscord = getMockDiscord();
+
         return [
-            [new User(getMockDiscord(), ['id' => '12345']), '12345'],
-            [new Channel(getMockDiscord(), ['id' => '12345']), '12345'],
-            [new Role(getMockDiscord(), ['id' => '12345']), '12345'],
+            [new User($mockDiscord, ['id' => '12345']), '12345'],
+            [new Channel($mockDiscord, ['id' => '12345']), '12345'],
+            [new Role($mockDiscord, ['id' => '12345']), '12345'],
             ['12345', '12345'],
             [null, null],
         ];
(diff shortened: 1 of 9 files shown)


davidcole1340 (dismissed):
thanks!

key2peace (approved):

SQKo:
These tests should be in separated workflow to reduce github actions time, since the result is not based the Discord API

tests/FunctionsTest.php
```diff
@@ -63,10 +81,147 @@ public function strlenProvider()
...
     /**
-     * @dataProvider strlenProvider
+     * @test
+     * @dataProvider mentionedProvider
      */
-    public function testPolyStrlen($expected, $string)
+    public function testMentioned($part, $attributes, $outcome): void
     {
-        $this->assertEquals($expected, poly_strlen($string));
+        $mockDiscord = getMockDiscord();
```
SQKo, line 89: I'm not sure why the same code is not used elsewhere

Exanlv: Not sure what you mean by this

SQKo: such as
```php
return [
	[new User(getMockDiscord(), ['id' => '12345']), '12345'],
	[new Channel(getMockDiscord(), ['id' => '12345']), '12345'],
	[new Role(getMockDiscord(), ['id' => '12345']), '12345'],
	['12345', '12345'],
	[null, null],
];
```

```
$mockDiscord = getMockDiscord();

return [
	[new User($mockDiscord, ['id' => '12345']), '12345'],
	[new Channel($mockDiscord, ['id' => '12345']), '12345'],
	[new Role($mockDiscord, ['id' => '12345']), '12345'],
	['12345', '12345'],
	[null, null],
];
```

Exanlv: Tbh I never thought about adding more logic then just a single return in a `dataProvider`

SQKo (approved):
Will think about the separate workflow later

PR #957, opened 2022-10-26
Guild->getBotPermission()

key2peace:


src/Discord/WebSockets/Events/GuildCreate.php
```diff
@@ -99,7 +99,7 @@ public function handle($data)
             });
         });
 
-        if ($this->discord->options['retrieveBans'] && $members[$this->discord->id]->getPermissions()->ban_members ?? true) {
+        if ($this->discord->options['retrieveBans'] && $guildPart->getBotPermissions()->ban_members ?? true) {
```
SQKo: Redundant calls to already cached members data from the local array

src/Discord/Parts/Guild/Guild.php
```diff
@@ -635,9 +635,7 @@ public function getVoiceRegions(): ExtendedPromiseInterface
      */
     public function createRole(array $data = [], ?string $reason = null): ExtendedPromiseInterface
     {
-        $botperms = $this->members->get('id', $this->discord->id)->getPermissions();
-
-        if ($botperms && ! $botperms->manage_roles) {
+        if (! $this->getBotPermissions()->manage_roles) {
```
SQKo, line 638: The `$botperms` was assigned to another variable because the getBotPermissions() can return null in case of missing cache, which prevented `null->manage_roles` error.

src/Discord/Parts/Guild/Guild.php
```diff
@@ -922,6 +922,16 @@ public function getAuditLog(array $options = []): ExtendedPromiseInterface
...
     }
 
+    /**
+     * Returns the bot's permissions in the guild.
+     *
+     * @return RolePermission|null
+     */
+    public function getBotPermissions(): ?RolePermission
+    {
+        return $this->members->get('id', $this->discord->id)->getPermissions();
```
SQKo, line 932: Should check if `$this->members->get()` returns null first

SQKo (approved):
untested

PR #980, opened 2022-11-13
Added fix for sub command group sub command autocompletes

CommandString:
From my 3 hours of digging around I have found that sub command group sub commands are never added the application_commands array when you begin listening for a command. This check uses similar logic to the anonymous checkCommand function but instead loops through sub_command_group options. Probably not the best way but it is working, my guess on a better fix would be to add sub command group sub commands to the application_commands array so it can be checked by checkCommand.

### DIFF
--- a/src/Discord/WebSockets/Events/InteractionCreate.php
+++ b/src/Discord/WebSockets/Events/InteractionCreate.php
@@ -69,7 +69,7 @@ class InteractionCreate extends Event
                             }
                         } elseif (! empty($option->focused)) {
                             return $command->suggest($interaction);
-                        } elseif ($option->type === Option::SUB_COMMAND_GROUP) {
+                        } elseif ($option->type === Option::SUB_COMMAND_GROUP || $option->type === Option::SUB_COMMAND) {
                             $checkSubCommandGroup = function ($options) use ($command, $interaction, &$checkSubCommandGroup) {
                                 foreach ($options as $option) {
                                     if (! empty($option->focused)) {
--- a/src/Discord/WebSockets/Events/InteractionCreate.php
+++ b/src/Discord/WebSockets/Events/InteractionCreate.php
@@ -12,6 +12,7 @@
 namespace Discord\WebSockets\Events;
 
 use Discord\InteractionType;
+use Discord\Parts\Interactions\Command\Option;
 use Discord\Parts\Interactions\Interaction;
 use Discord\Parts\User\User;
 use Discord\WebSockets\Event;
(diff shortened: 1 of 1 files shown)


SQKo:
I don't know your Issue, but v10 has breaking change that requires you to put the sub command name in the listenCommand argument.

```php
$discord->listenCommand(['cmd', 'sub'], function (Interaction $interaction) {
}, function (Interaction $interaction) use ($discord) {
    return [
        new Choice($discord, ['name' => 'one', 'value' => 1]),
        new Choice($discord, ['name' => 'two', 'value' => 2]),
        new Choice($discord, ['name' => 'second', 'value' => 3]),
    ];
});
```

gambar

CommandString:
> I don't know your Issue, but v10 (current dev-master) has breaking change that requires you to put the sub command name in the listenCommand argument, and it works fine.
> 
> ```
> $discord->listenCommand(['cmd', 'sub'], function (Interaction $interaction) {
> }, function (Interaction $interaction) use ($discord) {
>     return [
>         new Choice($discord, ['name' => 'one', 'value' => 1]),
>         new Choice($discord, ['name' => 'two', 'value' => 2]),
>         new Choice($discord, ['name' => 'second', 'value' => 3]),
>     ];
> });
> ```
> 
> gambar

🥲 that would explain why the name argument can be an array in the listenCommand method

SQKo:
I'd be glad if you explain the issue first with the repro code and command definition..

CommandString:
[Command Structure Definition](https://github.com/CommandString/san-andreas/blob/main/custom/Commands/Radio.php#L180-L227)

When try to use autocomplete on the **song** option it won't work unless I use the code I submitted in my PR. My guess is once I supply the sub commands everything will work.

CommandString:
I supplied sub commands to listen handler and now it works, so I guess my code isn't actually needed...woops.
*Also, not sure how or if I can retract this PR* 😅

SQKo:
> Well this worked for that one command but actually broke my other 2 sub commands. Adding more than 2 sub commands to the array list. No sub commands are handled.

You are just supposed to add more listener 

```php
$discord->listenCommand(['cmd', 'subgroup1'], ...)
$discord->listenCommand(['cmd', 'subgroup2'], ...)
$discord->listenCommand(['cmd', 'subgroup1', 'sub1'], ...)
$discord->listenCommand(['cmd', 'subgroup1', 'sub2'], ...)
```

It is designed that way to match the DiscordCommandClient (of message based)

CommandString:
Okay that worked! Thank you

PR #981, opened 2022-11-13
Fixed assignment issues after upgrading to PHP 8.1

mariopenterman:
Attempt to assign property "guild_id" on array


SQKo:
That doesn't seem right, can you show your code and the error log?

mariopenterman:
I made a big mistake during serializing of a part. This pull request can be closed.

PR #983, opened 2022-11-14
Cache compress

key2peace:


key2peace:
TODO: Move to (un)serializer

key2peace:
TODO:
Support for already compressed data even if cacheCompress has been disabled later by checking magic bytes (see RFC1950)
for this I will add a function in functions.php

SQKo (approved):
use squash

PR #993, opened 2022-11-23
fix messages thread stored as normal channel, fix ReactionRepository::delete() typehint

SQKo:
It was reported that threads are not properly handled, apparently messages events mistakenly treat threads as normal channels.

~~Untested.~~


SQKo:
Also changing type hinting for `ReactionRepository::delete($part)` argument to `Reaction`, since people are confused trying to put `MessageReaction` in it.

key2peace (approved):

PR #1000, opened 2022-12-01
Missing method "getBotPermissions" in Thread

xWildFire:
Found out that this method is missing in Thread and throwing me some errors on my app.


SQKo (approved):
thank

Issue #1004, opened 2022-12-04
message_count on null


xWildFire:
**Environment**
- PHP Version:
    - 8.1.13
- DiscordPHP Version:
    - dev-master (latest)

**Describe the bug**
https://github.com/discord-php/DiscordPHP/blob/53729d5a338bf0597b07c834796b3791704d7f8f/src/Discord/WebSockets/Events/MessageDelete.php#L58

This line is sometimes causing error cause of $parent being null. Idk how it should correctly work so im just reporting it.

PR #1005, opened 2022-12-05
Only reduce message_count if it's a thread

key2peace:
#1004

### DIFF
--- a/src/Discord/WebSockets/Events/MessageDelete.php
+++ b/src/Discord/WebSockets/Events/MessageDelete.php
@@ -29,7 +29,7 @@ class MessageDelete extends Event
      */
     public function handle($data)
     {
-        $messagePart = $parent = null;
+        $messagePart = null;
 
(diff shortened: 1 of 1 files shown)


SQKo:
No, it wont work like this in forum, i was oversimplyfing this code but overlooked normal channels where it'd be always null
https://github.com/discord-php/DiscordPHP/pull/993/commits/a8a273fb56939464d2d589d8bddca33a9770dfbe#diff-32749737b5ed6d242a013418143adc92a0a33f2182a9553d8f50b4f0a8d667c5L44

SQKo (approved):
I guess we back to instanceof again,
also returned back the code order because it rely on the cache $messagePart being pulled

SQKo (changes requested):
Hold on, the count is now wrong for normal threads

SQKo (approved):
Fixed

PR #1017, opened 2022-12-13
Add Application::role_connections_verification_url

SQKo:
Reference: https://github.com/discord/discord-api-docs/pull/5668

Not sure if the role connection object need to be made up as Part, I have yet to see it used in the Bot rather than an application's endpoint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Before you start using this Library, you **need** to know how PHP works, you nee
 
 #### Recommended Extensions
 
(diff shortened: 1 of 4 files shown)


key2peace:
> Not sure if the role connection object need to be made up as Part, I have yet to see it used in the Bot rather than an application's endpoint.

yes it does, see [Application Role Connection Metadata](https://discord.com/developers/docs/resources/application-role-connection-metadata)

SQKo:
TODO: Add https://github.com/discord/discord-api-docs/pull/5771 later since it may conflict with #1018

PR #1021, opened 2022-12-13
Add default_thread_rate_limit_per_user in Channel creatable attributes

SQKo:
https://github.com/discord/discord-api-docs/pull/5606

`default_thread_rate_limit_per_user` is for setting the thread default `rate_limit_per_user` which exists on the text channel and forum channel.

~~Since right now `flags` only relevant to forums, i only added flags for forum.~~

Also deprecated Channel flag pinned and moved it to Thread class


key2peace (dismissed):

SQKo:
Removed flags from creatable attributes since it's not working.

key2peace (approved):

PR #1037, opened 2023-01-10
added scheme checks in embed.php

PeanutNL:
Follow up on #1036
Added checks on the urls on the embeds

### DIFF
--- a/src/Discord/Parts/Embed/Embed.php
+++ b/src/Discord/Parts/Embed/Embed.php
@@ -368,17 +368,35 @@ class Embed extends Part
             throw new \LengthException('Author name can not be longer than 256 characters.');
         } elseif ($this->exceedsOverallLimit($length)) {
             throw new \LengthException('Embed text values collectively can not exceed than 6000 characters');
-        } else if ($iconurl != null && preg_match('/^(http|https|attachment):\/\//i', $iconurl) != 1) {
-            throw new \InvalidArgumentException("Iconurl scheme only supports http(s) and attachments");
-        } else if ($url != null && preg_match('/^(http|https|attachment):\/\//i', $url) != 1) {
-            throw new \InvalidArgumentException("Url scheme only supports http(s) and attachments");
-        } else {
-            $this->author = [
-                'name' => $name,
-                'icon_url' => $iconurl,
-                'url' => $url,
-            ];
         }
+
+        if ($iconurl !== null) {
+            $iconurl = filter_var($iconurl, FILTER_VALIDATE_URL);
+            if ($iconurl === false) {
+                throw new \InvalidArgumentException("Invalid url");
+            }
+            $iconurlscheme = parse_url($iconurl, PHP_URL_SCHEME);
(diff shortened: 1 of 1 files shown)


PeanutNL:
Replaced the preg_match checks!

Used filter_var to make sure its an URL.
used parse_url to make sure it has a correct scheme.

key2peace (approved):

key2peace (dismissed):

src/Discord/Parts/Embed/Embed.php
```diff
@@ -469,6 +504,26 @@ public function setURL(string $url): self
...
     }
 
+    /**
+     * Checks to see if an url is valid to be used in embeds.
+     *
+     * @param ?string $url
+     *
+     * @return bool
+     */
+    protected function checkEmbedUrl(?string $url, array $allowed = ['http', 'https', 'attachment']): bool
```
Exanlv, line 514: The name `checkEmbedUrl` doesn't really indicate what it's checking. `validateUrlScheme` or something like that would be more descriptive.

src/Discord/Parts/Embed/Embed.php
```diff
@@ -469,6 +504,26 @@ public function setURL(string $url): self
...
+    {
+        if (null === $url) return true;
```
Exanlv, line 516: `null` isn't using a valid url scheme. Instead of returning `true` to allow it to pass through the check, it should not be going into this check at all.

Instead of `if (! $this->checkEmbedUrl($url))` in other functions you should use `if (is_null($url) || ! $this->checkEmbedUrl($url))`

src/Discord/Parts/Embed/Embed.php
```diff
@@ -469,6 +504,26 @@ public function setURL(string $url): self
...
+
+        if ($url === '') return false;
```
Exanlv, line 518: An early return like this is unnecessary. If you attempt to parse the url scheme from an empty string it'd also return `false`

Exanlv (changes requested):

src/Discord/Parts/Embed/Embed.php
```diff
@@ -469,6 +504,26 @@ public function setURL(string $url): self
...
+
+        if (filter_var($url, FILTER_VALIDATE_URL) === false) return false;
+
+        if (! in_array(parse_url($url, PHP_URL_SCHEME), $allowed) return false;
```
Exanlv, line 522: This is the only required check in this function (tho there is a syntax error) the entire function can be simplified to the following

```suggestion
        return in_array(parse_url($url, PHP_URL_SCHEME), $allowed);
```

src/Discord/Parts/Embed/Embed.php
Exanlv, line 514: **Note:**
Some internal discussion is still going on about how this function should work. I'll resolve this comment after we get to something conclusive

Exanlv (dismissed):

key2peace (approved):

PR #1045, opened 2023-01-22
Remove useless forced dependency (psr/simple-cache)

Divi:
Hello!

The dependency `psr/simple-cache` is locked at a very old version (~2017, 1.x, now the 3.x exists, the locked version blocks every recent library that we want to add). This dependency should not be required, only suggested because it's used only for resolving the cache configuration parameter.

I removed it from composer.json file and handled the case where the dependency does not exist.


key2peace:
as described on discord, this is not useless/unused but part of the cacheInterface. #1046 is what should be done instead. closing

PR #1057, opened 2023-02-05
Change 'ready' to 'init'

key2peace:


CommandString (dismissed):

src/Discord/Discord.php
```diff
@@ -1121,17 +1121,20 @@ protected function send(array $data, bool $force = false): void
...
+        if ($this->emittedInit) {
             return false;
         }
 
         $this->logger->info('client is ready');
-        $this->emit('ready', [$this]);
+        $this->emit('init', [$this]);
+
+        /* deprecated */
+        $this->emit('ready', [$this]); // deprecated
```
Exanlv, line 1137: User has no way of knowing this is deprecated, there should be a notice of sorts.

Doesn't look like `evenement` has a way of marking events as deprecated sadly, something along these lines could work however.

```php
if (isset($this->listeners['ready']) {
    $this->logger->info('The \'ready\' event is deprecated and will be removed in a future version of DiscordPHP. Please use \'init\' instead.')
}
```

Exanlv (changes requested):

src/Discord/Discord.php
key2peace: applied after fixing a few typo's. hope ya like it :)

PR #1069, opened 2023-03-14
Fixed Emoji __toString being malformed

CommandString:
As of now the `Emoji::__toString()` method returned `<name:id>` when it should be returning `<:name:id>`


Exanlv (approved):
Verified syntax by escaping an emoji in discord. Syntax of new implementation matches this.

I don't think this is a breaking change, (though maybe because its 1 more character you could consider it that?), but maybe throw it in contrib chat before merging anyway

PR #1087, opened 2023-04-11
Make deleteMessages for threads & channels uniform

ellisonpatterson:
I know there was a todo to make a trait, but this will suffice for now.

The threads `deleteMessages` didn't allow for `Traversable` objects to work, so `Collections` weren't working.

This fixes that and now I can use `deleteMessages` interchangeably on Interaction responses without having to check if it was triggered in a Channel or Thread.

### DIFF
--- a/src/Discord/Parts/Thread/Thread.php
+++ b/src/Discord/Parts/Thread/Thread.php
@@ -16,6 +16,7 @@ use Discord\Builders\MessageBuilder;
 use Discord\Helpers\Collection;
 use Discord\Helpers\Deferred;
 use Discord\Http\Endpoint;
+use Discord\Http\Exceptions\NoPermissionsException;
 use Discord\Parts\Channel\Channel;
 use Discord\Parts\Channel\Message;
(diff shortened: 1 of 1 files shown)


src/Discord/Parts/Thread/Thread.php
```diff
@@ -476,7 +479,7 @@ public function getPinnedMessages(): ExtendedPromiseInterface
      *
      * @link https://discord.com/developers/docs/resources/channel#bulk-delete-messages
      *
-     * @param array       $messages
+     * @param array|Traversable $messages An array of messages to delete.
      * @param string|null $reason   Reason for Audit Log (only for bulk messages).
```
Exanlv, line 483:
```suggestion
     * @param array|Traversable $messages An array of messages to delete.
     * @param string|null       $reason   Reason for Audit Log (only for bulk messages).
```
Running the linting should also fix this

Exanlv (changes requested):

Exanlv (approved):

PR #1116, opened 2023-05-07
Add AutoModeration Action metadata, and the custom_message field

SQKo:
~~Metadata will now have its own class so we have less boilerplate mutator in each Part~~, and this allows typehinting the metadata

Also added custom_message metadata for auto moderation action
https://github.com/discord/discord-api-docs/pull/5955


SQKo:
I changed my mind on a new class, we should be implementing guard property for the Part fillables and just create a metadata Part that has no guard.

key2peace (approved):

PR #1121, opened 2023-05-18
Revert "Static method to get loop"

key2peace:
Reverts discord-php/DiscordPHP#1120


SQKo:
As a note in future: I know the factory was deprecated, I did this intentionally long ago to fix a weird issue and compatibility with those using older dependencies and various use of loop options and php < 8.1

As soon we have minimum 8.1, this will be no longer issue. But at the same time the options would not be actually needed anymore in future (no more `$discord(['loop' => $loop]);`) and another fun fact is that we would not need the `$discord->run()` anymore (You can try it already if you are on php 8.1+ with latest react loop lib + remove the $discord->run()` from your code)

SQKo (approved):

PR #1143, opened 2023-06-30
generateSnowflake($timestamp)

key2peace:
useful function I might need later in the code :)


SQKo:
Sorry, I don't think it's good to add something misleading from the specification/standard. those sequences are not random, people gonna use this function thinking it works the same way but it's not.

PR #1145, opened 2023-07-01
update getFieldsAttribute()

key2peace:
this was the only function actually using Collections. all other fields related functions (addField, setFieldAttribute) use plain array manipulation. another thing is, this field is optional and should therefor return null if the attribute is not set.


SQKo:
Other fields aren't actually null due to `attributesHelper()` creating the object, so it wouldn't be consistent.
I must've mistaken when documenting the @property

In future if we move out the builder functions to separate class then we can consider making the Part attributes actually optional

PR #1147, opened 2023-07-04
Add $member->kick() and improve $member->ban()

key2peace:
Add $member->kick(optional audit log reason), with permission checks


SQKo:
There is already `$guild->members->kick($member, $reason)`, if you insist to add, please just call the same function instead of duplicating whole codes.

SQKo (approved):
changed to use repository but permission checks are kept

PR #1156, opened 2023-07-16
make Part::$http property dynamic only when needed

SQKo:
Not All Part classes needed the http client stored as property, this PR makes it optional that only the Part classes that needs the http client. While this add overhead to the getter, this in hope reduces memory by removing unused property.

So far there is no significance difference, testing it with D.PHP bot for atleast 24 hours.


SQKo:
There was no significance change, so I'm closing this

PR #1157, opened 2023-07-17
Promise v3

key2peace:


src/Discord/Discord.php
```diff
@@ -42,15 +42,15 @@
...
 use Discord\Helpers\RegisteredCommand;
 use Discord\Http\Drivers\React;
 use Discord\Http\Endpoint;
 use Evenement\EventEmitterTrait;
 use Monolog\Formatter\LineFormatter;
 use Psr\Log\LoggerInterface;
 use React\Cache\ArrayCache;
-use React\Promise\ExtendedPromiseInterface;
+use React\Promise\Deferred;
+use React\Promise\PromiseInterface;
```
key2peace, line 53: oi, duplicate, will fix

2colours:
@key2peace what is the state of this PR? Being stuck with v2 Promises (thanks, Composer, for not handling multiple versions of one package btw) means even something as banal as `set_rejection_handler` cannot be used.

SQKo:
> @key2peace what is the state of this PR? Being stuck with v2 Promises (thanks, Composer, for not handling multiple versions of one package btw) means even something as banal as `set_rejection_handler` cannot be used.

It wont be in v10, many dependencies still stuck with v2 promises
As well compatibility issue with future php 8.4 version, which they'll promise to support by reactphp v3

about set_rejection_handle, Just be careful with your promise codes for now
always use `then()` followed by argumentless and empty `done()` so you can refactor your code easier in future

2colours:
@SQKo DiscordPHP should at the very least handle the rejection path with a logger, with a log level >= warning. There is no reason to not do that since afterwards nobody could even catch the rejection currently. I think that's way better than leaving it up to the customers to develop some quasi framework that minimizes the boilerplate for handling this for each and every registered command.

I'm already looking into the code; it's just not easy to navigate around with something you have only been actively using for like 2 months...

2colours:
see https://github.com/discord-php/DiscordPHP/pull/1208

This is anything but exhaustive, it just covers my most obvious use case.

SQKo:
> @SQKo DiscordPHP should at the very least handle the rejection path with a logger, with a log level >= warning. There is no reason to not do that since afterwards nobody could even catch the rejection currently. I think that's way better than leaving it up to the consumers to develop some quasi framework that minimizes the boilerplate for handling this for each and every registered command.
> 
> I'm already looking into the code; it's just not easy to navigate around with something you have only been actively using for like 2 months...

We know about it already, unfortunately the issue we are facing also caused by dependencies, our code is fine when web socket is not connected yet..

Regarding to your suggestion, we already considered about it, but it is wrong to log everything even those promise rejections already handled/caught by users and not respecting users configured Logger.
It is somewhat already implemented but requires your code to be wrapped inside a coroutine yields

https://github.com/discord-php/DiscordPHP/blob/ac16fdf6e9a9d20003edbca734c386ddcf616eda/src/Discord/Discord.php#L793-L799
notice why it is logged only in websocket events because it only gets bugged after websocket connected...

valzargaming:
wyrihaximus/react-cache-redis is currently locked to `^3.0 || >=4.0 <4.4` and we are required to use at least `^4.5.0` to include `react/promise ^3.1`. This is considered to be a core feature of the v10 library to support external caching and must be tested before this can be included as part of a new release.

valzargaming (approved):

PR #1198, opened 2024-01-23
Replace is_resource usage with !== false

valzargaming:
See https://wiki.php.net/rfc/resource_to_object_conversion


SQKo:
on hold because it depends on the reactphp library side & release of php 8.4

PR #1199, opened 2024-02-02
feat : string possibility with interact

Refaltor77:
Subtlety, but which allows a tremendous time-saving on large bot projects with this API :)


src/Discord/Parts/Channel/Channel.php
```diff
@@ -262,6 +262,11 @@ protected function getRecipientsAttribute(): Collection
         return $recipients;
     }
 
+
+
+
+
+
```
Nirbose, line 269: why ?

src/Discord/Parts/Channel/Channel.php
```diff
@@ -350,6 +355,25 @@ public function setPermissions(Part $part, array $allow = [], array $deny = [],
...
+        if ($this->name === $name) {
+            return reject(new InvalidOverwriteException('The name is already set in the channel.'));
+        }
+
+        $this->name = $name;
+        return $this->http->patch(Endpoint::bind(Endpoint::CHANNEL, $this->id),[
+            'name' => $this->name
+        ]);
+    }
+
```
key2peace, line 376: This should be in a separate PR

SQKo: I disagree with this, there is already `$channel->name` setter and `save()`
we cant let adding every function for changing every simple attribute like this

people would make multiple different http calls just to update a channel information

SQKo (dismissed):
please uncommit the rename() method, besides the reasons i mentioned in the comment above, it is also out of the PR scope

valzargaming (approved):

src/Discord/Parts/Channel/Channel.php
Refaltor77: why not 🌚

PR #1207, opened 2024-03-16
Update Application

key2peace:
Added missing attributes and placed them in the order of api docs

### DIFF
--- a/src/Discord/Parts/OAuth/Application.php
+++ b/src/Discord/Parts/OAuth/Application.php
@@ -167,12 +167,12 @@ class Application extends Part
      */
     protected function getBotAttribute(): ?User
     {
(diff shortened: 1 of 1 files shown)


SQKo:
By the way to be noted for other part class, do not sort `fillable` by Discord API docs but:

1. ID go first
2. Repositories Go Last
3. Equals/close to 1:1 to the actual payload

The first and second was important in Channel and Guild, e.g. permission, it was bugged once since the guild id is missing and the permission was stored for null channel/guild id. It really that matters.
For 3rd, it is for parsing json speed.

The orders in discord api are usually inserted by whoever made PR in the docs api, thus cannot be used as as `fillable` order, PHPdocs attributes are fine though

PR #1231, opened 2024-05-27
Adds toCommand helper function to CommandBuilder

valzargaming:
We've previously recommended the previous flow for creating new commands using the command builder
```php
$discord->application->commands->save(
    $discord->application->commands->create($commandbuilder);
);
```
With this tweak, we can cut out an extra call to commands->create
```php
$discord->application->commands->save($commandbuilder->toCommand($discord));
```

### DIFF
--- a/src/Discord/Builders/CommandBuilder.php
+++ b/src/Discord/Builders/CommandBuilder.php
@@ -11,6 +11,7 @@
 
 namespace Discord\Builders;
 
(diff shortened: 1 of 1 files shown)


SQKo:
Would be better to override the CommandRepository save method() to accept command builder because the created part is determined by the repository create(). instead of always forcing the commands as new.

PR #1237, opened 2024-06-23
Enhancement to addRole and removeRole methods for Member class

valzargaming:
This improvement aims to optimize the Member class methods, specifically addRole and removeRole, making them more versatile and aligned with expected behavior, thereby enhancing the overall usability of the library. Currently, due to their use of PUT and DELETE endpoints, these methods do not return an updated member part. The proposed change introduces the ability to pass true as the third parameter, enabling the use of array_merge or array_diff operations and facilitating the return of an updated member part by calling the PATCH endpoint instead.

This enhancement grants additional functionality by allowing the methods to return a resolved member, thereby enhancing the library’s capabilities without compromising its existing structure. Doing so would also add better support for chaining of promises in this context.

It's important to note that chaining multiple addRole or removeRole promises together will always be slower and result in additional API calls where it would otherwise be better to make a single call to setRoles. This change is not intended to replace or undermine the intended usage of setRole, but reduce the amount of code required for a user of the library to effectively utilize these functions.

### DIFF
--- a/src/Discord/Parts/User/Member.php
+++ b/src/Discord/Parts/User/Member.php
@@ -288,11 +288,11 @@ class Member extends Part implements Stringable
                     return $this;
                 })
             : $this->http->put(Endpoint::bind(Endpoint::GUILD_MEMBER_ROLE, $this->guild_id, $this->id, $role), null, $headers)
-            ->then(function () use ($role) {
-                if (! in_array($role, $this->attributes['roles'])) {
-                    $this->attributes['roles'][] = $role;
-                }
-            });
+                ->then(function () use ($role) {
+                    if (! in_array($role, $this->attributes['roles'])) {
+                        $this->attributes['roles'][] = $role;
+                    }
+                });
     }
 
     /**
--- a/src/Discord/Parts/User/Member.php
+++ b/src/Discord/Parts/User/Member.php
@@ -249,13 +249,14 @@ class Member extends Part implements Stringable
      *
      * @param Role|string $role   The role to add to the member.
      * @param string|null $reason Reason for Audit Log.
+     * @param bool|null   $set    Whether to set the roles using PATCH instead of PUT and return the updated member part on resolved promise.
      *
      * @throws \RuntimeException
      * @throws NoPermissionsException Missing manage_roles permission.
      *
-     * @return ExtendedPromiseInterface
+     * @return ExtendedPromiseInterface|ExtendedPromiseInterface<static>
      */
-    public function addRole($role, ?string $reason = null): ExtendedPromiseInterface
+    public function addRole($role, ?string $reason = null, ?bool $set = false): ExtendedPromiseInterface
     {
         if ($role instanceof Role) {
             $role = $role->id;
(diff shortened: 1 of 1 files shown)


valzargaming:
I've been looking for a flexible userland solution that could effectively be used instead which would obtain the same result without this PR being merged. This is what I've come up so far:
```php
/**
 * Removes specified roles from a member.
 *
 * @param Member $member The member object from which to remove the roles.
 * @param Collection<Role>|array<string|int>|Role|string|int $roles An array of role IDs to be removed.
 * @param bool $patch Determines whether to use patch mode or not. If true, the member's roles will be updated using setRoles method. If false, the member's roles will be updated using removeRole method.
 * @return PromiseInterface<Member> A promise that resolves to the updated member object.
 */
public function removeRoles(Member $member, Collection|array|Role|string|int $roles, bool $patch = true): PromiseInterface
{
    if (! $roles) return \React\Promise\resolve($member);
    if (! is_array($roles) && ! $roles instanceof Collection) $roles = [$roles];
    $role_ids = [];
    if ($roles instanceof Collection && $roles->first() instanceof Role) $role_ids = $roles->map(fn($role) => $role->id)->toArray();
    elseif ($roles instanceof Role) $role_ids[] = $roles->id;
    else foreach ($roles as $role) if ($member->guild->roles->get('id', $role)) $role_ids[] = "$role";
    if (! $role_ids) {
        $this->logger->warning('Roles not found for removeRoles: ' . json_encode($roles));
        return \React\Promise\resolve($member);
    }
    if (($new_roles = $member->roles->filter(function (Role $role) use ($roles) { return ! in_array($role->id, $roles); })) === $roles) return \React\Promise\resolve($member);
    return $patch
        ? $member->setRoles(array_diff($member->roles->toArray(), $new_roles))
        : \React\Promise\all(array_map(fn($role) => $member->removeRole($role->id), $roles))
            ->then(function() use ($member) {
                return $member;
            });
}
```

valzargaming:
Below are the new recommended userland implementations of addRoles and removeRoles based on the recent changes to Collections.
```php
/**
 * Removes specified roles from a member.
 *
 * @param Member                                                  $member The member object from which the roles will be removed.
 * @param Collection<Role>|array<Role|string|int>|Role|string|int $roles  An array of role IDs to be removed.
 * @param bool                                                    $patch  Determines whether to use patch mode or not. If true, the member's roles will be updated using setRoles method. If false, the member's roles will be updated using removeRole method.
 * @return PromiseInterface<Member>
 */
public function removeRoles(
    Member                           $member,
    Collection|array|Role|string|int $roles,
    bool                             $patch = true
): PromiseInterface
{
    if (! $role_ids = array_filter(self::__rolesToIdArray($roles), fn($role_id) => $member->roles->has($role_id))) return resolve($member);
    return $patch
        ? ((($new_roles = $member->roles->filter(static fn(Role $role) =>
            ! in_array($role->id, $role_ids))) !== $member->roles)
                ? $member->setRoles($new_roles->toArray())
                : resolve($member))
        : array_reduce($role_ids, fn($promise, $role_id) =>
            $promise->then(static fn() => $member->removeRole($role_id)),
            $member->removeRole(array_shift($role_ids)))
                ->then(static fn() => $member->guild->members->get('id', $member->id));
}
```
```php
/**
 * Adds specified roles to a member.
 *
 * @param Member                                                  $member The member object to which the roles will be added.
 * @param Collection<Role>|array<Role|string|int>|Role|string|int $roles  An array of role IDs to be added.
 * @param bool                                                    $patch  Determines whether to use patch mode or not. If true, the member's roles will be updated using setRoles method. If false, the member's roles will be updated using addRole method.
 * @return PromiseInterface<Member>
 */
public function addRoles(
    Member                           $member,
    Collection|array|Role|string|int $roles,
    bool                             $patch = true
): PromiseInterface
{
    if (! $role_ids = array_filter(self::__rolesToIdArray($roles), fn($role_id) => $member->roles->has($role_id))) return resolve($member);
    return $patch
        ? $member->setRoles(
            array_merge($member->roles
                ->map(fn($role) => $role->id)
                ->values(),    
            $role_ids))
        : array_reduce($role_ids, fn($promise, $role_id) =>
                $promise->then(static fn() => $member->addRole($role_id)),
                $member->addRole(array_shift($role_ids)))
                    ->then(static fn() => $member->guild->members->get('id', $member->id));
}
```

PR #1251, opened 2024-09-21
Implement volume calculations to dB for use with ffmpeg

brunofunnie:
There was no effective volume implementation, so I've added this linear calculation solution.


valzargaming:
Looks good to me, but I've not used it myself, so I've tagged others for additional review who I know have.

valzargaming (approved):

PR #1256, opened 2024-09-27
Overwrite repository->save for CommandBuilder

valzargaming:
@see #1231


valzargaming:
Closing this as both stale and not planned. We now have `abstract class Builder` with `public static function fromPart(Part $part): self` and a similar feature is planned for Parts.

PR #1263, opened 2024-11-19
Update offsetExists Function to Use Async Cache Check

valzargaming:
# Description
This PR updates the `offsetExists` function in the `AbstractRepository` class to enhance functionality by integrating an asynchronous cache check. The previous implementation used a direct parent call for `offsetExists`. The new implementation leverages the `cache` component, ensuring a more accurate check of the cache state.

# Changes Made

- Updated `offsetExists` from: `return parent::offsetExists($offset)` to `return await($this->cache->has($offset))`.
- This change ensures the repository adheres to an asynchronous design pattern and integrates seamlessly with the `cache` system.

# Benefits
- Adds forward compatibility for the previously deprecated `offsetExists` function, allowing it to remain functional with the new `cache` system.

### DIFF
--- a/src/Discord/Repository/AbstractRepository.php
+++ b/src/Discord/Repository/AbstractRepository.php
@@ -24,6 +24,8 @@ use Traversable;
 use WeakReference;
 
 use function Discord\nowait;
+
+use function React\Async\await;
 use function React\Promise\reject;
 use function React\Promise\resolve;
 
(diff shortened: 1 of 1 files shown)


bwoebi:
I'm randomly seeing this - and would like to note that this will be quite the breaking change. At least unless you make all discord event callbacks async too.
Currently awaiting will ultimately stall ratchet in its callbacks, causing the current pending websocket packets to be dequeued after the callback possibly causing out-of-order callback delivery. (and this will cause the connection to become invalid for example)

valzargaming:
I'm more than a little disappointed that you're the only person to point that out.

PR #1267, opened 2024-12-24
Implement modularity Collection class with 

valzargaming:
The main singleton object now includes a new `getCollectionClass` method to centralize the collection class logic. Parts of the library that previously used the hardcoded Collection class now dynamically retrieve the appropriate class via the getCollectionClass method.


valzargaming:
The only thing that still needs to be done is reworking the AbstractRepository class to support this change.

valzargaming:
Work on this will be done on the v11 branch

PR #1272, opened 2024-12-30
Deafen and Undeafen members

valzargaming:
Adds `deafenMember` and `undeafenMember` functionality to voice channels.

### DIFF
--- a/src/Discord/Parts/Channel/Channel.php
+++ b/src/Discord/Parts/Channel/Channel.php
@@ -571,6 +571,76 @@ class Channel extends Part implements Stringable
     }
 
     /**
+     * Deafens a member on a voice channel.
+     *
+     * @param Member|string $member The member to deafen. (either a Member part or the member ID)
+     * @param string|null   $reason Reason for Audit Log.
+     *
+     * @throws \RuntimeException
+     * @throws NoPermissionsException Missing deafen_members permission.
+     *
+     * @return PromiseInterface
+     */
+    public function deafenMember($member, ?string $reason = null): PromiseInterface
+    {
+        if (! $this->isVoiceBased()) {
+            return reject(new \RuntimeException('You cannot deafen a member in a text channel.'));
+        }
+
+        if ($botperms = $this->getBotPermissions()) {
+            if (! $botperms->deafen_members) {
+                return reject(new NoPermissionsException("You do not have permission to deafen members in the channel {$this->id}."));
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


mikield:
One interesting though that came to me in this PR:

Can we have a context-full resources?
What I mean that, it would be awesome to have something like `$channel->members->deafen(ID)` instead of calling `$channel->deafenMember()`.

We can return another type of `MembersRepostory` for voice channels.
So `VoiceChannel::memebrs` can be `VoiceMembersRepository` where we can perform voice-specific actions on members.

valzargaming:
These two functions just update the member's properties and send them back to Discord to overwrite them. You could get this same effect by calling `$member->deafen = true` and following up with `$guild->members->save($member)`, so it could be argued (correctly so) that this PR is inappropriate to begin with, and the logic should be updated to merely update the property and leave it u[ to the user to decide when the part should be updated. This is the same way you'd rename a channel, update the property and save it to the guild's channel repository.

I know right now it seems like there's a missing feature here, but the reality is some functions like un/deafening members don't actually have API endpoints associated with them. While I agree that we do need a better way to access the functionality from out of this specific context, I don't think adding aliases or unnecessary calls to the PATCH endpoint is the way to go about it. I've been considering adding traits to the classes that specifically handle updating parts to prep them for saving, but nothing is set in stone yet as I'm not sure how we would futureproof it.

valzargaming:
I'm open for any suggestions on how to accomplish this. Looking things through some more I still think it makes more sense to scrap this PR completely.

Ideally a helper method should be made that would return a PromiseInterface so it can `reject` on a permissions error and `resolve($this)` on success. We'd need to set up something like `setDeafAttribute`, but the biggest roadblock is that the current implementation is hijacking the `__set` magic method which only allows for a void return... That brings us back to my original idea, creating new functions with PromiseInterface returns, and the original problem.

valzargaming:
The docs are surprisingly scarce as to how un/deafening and un/muting should work. For deafening, the only requirements are the bot having the DEAFEN_MEMBERS bitwise permission and for the member to currently be in a voice channel, necessitating the need to call the function within the context of a channel. Muting members is identical except that the bot is allowed to mute itself. The only way we're able to implement this properly at this time is with the original suggestion, else the Member part would need to accept a channel as a parameter, which could cause further confusion as it'd be possible to pass a channel the member isn't already in, and would break the library's pattern.s

valzargaming:
This should be production ready now

PR #1275, opened 2025-01-14
Update SelectMenu.php

mikield:
Just a small fix for auto-responding in a Select Menu interaction.

When I have an interaction that I have already responded in listener, `SelectMenu` code doesn't care about it, it calls `$ack` function anyway.

I do not change how it is called, but I think it should check if an interaction was responded before, before trying to `acknowledge` it.


valzargaming:
The same could be done for Button, but I don't see that isResponded is ever actually used except in the case of getOriginalResponse, respond, and getFollowUpMessage. It doesn't seem to be very reliable if acknowledge doesn't set that flag either, so even with this PR I don't think it'd do what's intended.

valzargaming:
I'll push a new patch release today

PR #1279, opened 2025-01-18
fix: PollResults to correctly map to an array of PollAnswerCount

sweptsquash:
While working on a small integration utilizing Channel Polls, I encountered an issue where retrieving the `answer_counts` unexpectedly returned a singular entity instead of an array of `PollAnswerCount` objects. This behavior was inconsistent with the expected structure and functionality.

This PR addresses the issue by ensuring that the `answer_counts` object is properly mapped into an array of `PollAnswerCount` objects, aligning with the intended design and improving data handling.

```
Discord\Parts\Channel\Poll\PollAnswerCount^ {#3876
  #http: Discord\Http\Http^ {#3813}
  #factory: Discord\Factory\Factory^ {#3064}
  #discord: Discord\Discord^ {#3856}
  +scriptData: null
  #fillable: array:3 [
    0 => "id"
    1 => "count"
    2 => "me_voted"
  ]
  #attributes: []
  #visible: []
  #hidden: []
  #repositories: []
  #repositories_cache: []
  +created: true
  id: null
  count: null
  me_voted: null
}
```

### DIFF
--- a/src/Discord/Parts/Channel/Poll/PollResults.php
+++ b/src/Discord/Parts/Channel/Poll/PollResults.php
@@ -22,7 +22,7 @@ use Discord\Parts\Part;
  *
  * @since 10.0.0
  *
- * @property boolean            $is_finalized   Whether the votes have been precisely counted
+ * @property boolean                       $is_finalized   Whether the votes have been precisely counted
  * @property Collection|PollAnswerCount[]  $answer_counts  The counts for each answer
  */
 class PollResults extends Part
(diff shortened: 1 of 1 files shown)


sweptsquash:
@valzargaming Please let me know if you're happy with me making use of the collection helper here or if you would prefer me to use an alternative method.

valzargaming:
I'm a bit confused why you're creating a new Collection object only to use it's ArrayAccess properties with ->push and then converting it back to an array. Why not just use `$var []= $value`?

sweptsquash:
> I'm a bit confused why you're creating a new Collection object only to use it's ArrayAccess properties with ->push and then converting it back to an array. Why not just use `$var []= $value` or leave it as a Collection?

Fair point, I've pushed a commit to bring this more inline with similar methods and now return the `CollectionInterface`

valzargaming (dismissed):

valzargaming (approved):

PR #1288, opened 2025-01-27
Dockerized development

SQKo:
This adds a simple dockerfile and compose, with everything in the bot folder mounted (except in dockerignore), using php8.1 CLI (alpine for small size), xdebug included.

For the PHP version will be using the minimum required for the version, so make sure to update this when composer.json requirement is bumped up too.

By default it will run the `bot.php` in the project directory, no entrypoint as I believe people in development usually go to the docker terminal and re-execute the script manually.

Simply override the `BOTDIR` and `BOTSCRIPT` environment variable to use another desired path.


Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+ENV BOTDIR="/root/discord-php"
+ENV BOTSCRIPT="./bot.php"
+LABEL version="10"
+WORKDIR ${BOTDIR}
+COPY . ${BOTDIR}
+
+COPY --from=composer /usr/bin/composer /usr/bin/composer
+RUN composer install --no-cache --no-interaction
+
+CMD php ${BOTSCRIPT}
```
Exanlv, line 21:
```suggestion
CMD php ${BOTSCRIPT}

```

docker-compose.yml
```diff
@@ -0,0 +1,8 @@
+services:
+  bot:
+    build: .
+    volumes:
+     - .:/root/discord-php
+volumes:
+  bot:
+    driver: local
```
Exanlv, line 8:
```suggestion
    driver: local

```
End files with empty lines :^)

Exanlv:
You should add `docker-compose.override.yml` to `.gitignore` as well, in case someone needs a local override, they can't accidentally commit it

Exanlv (approved):

PR #1291, opened 2025-02-07
Move Collection to its own repository

valzargaming:


Exanlv:
Not gonna lie, I don't see how this is beneficial. 

Just creates more overhead to maintain it separately?

valzargaming:
Several of the existing functions are outdated to be backwards compatible with old PHP versions. By moving it to a new repository we can introduce releases with their own PHP minimum versions and composer will automatically update to the end-user's most compatible up-to-date version without needing to force everybody to use PHP 8.4.

valzargaming:
This actually makes maintaining the classes much easier as we can choose to drop backwards compatibility at any time in the future without needing to worry about rewriting functions at the same time, and it removes the extra effort from end-users as composer.json will handle choosing the appropriate version.

Both branches of the new repository will respect the `CollectionInterface` but have their own PHP version specific ways of doing things. The 8.4 branch contains the 3 additional array methods that were introduces with 8.4, and if there is enough of a demand for it we can add our own 8.0-specific implementation for it on the 8.0 branch. It's a win-win and will save us time in the future.

Exanlv:
I don't think its a good idea to reuse the namespace. Feels like youre just waiting for the autoloader to pull the class from the wrong place to me

valzargaming:
I'm not sure I follow. Why would we want to change the namespace on an existing class within the library? The only difference is we've moved it to its own repository. I don't see how the autoloader would come into play here unless someone using the library explicitly tried to reuse our namespaces.

valzargaming:
We've decided against this approach for the time, but I will likely continue to maintain the external repository regardless so it cna be used by other projects.

PR #1294, opened 2025-02-21
🚧 Initial Discord message components v2 implementation

Log1x:
# 🚧 Work in progress.

Completely blind & untested. Should be able to try it later today.

Edit: everything appears to be functional.

Screenshot

### DIFF
--- a/src/Discord/Builders/Components/Container.php
+++ b/src/Discord/Builders/Components/Container.php
@@ -24,7 +24,7 @@ use Discord\Builders\Components\Separator;
  *
  * @link https://discord.com/developers/docs/interactions/message-components#container
  *
- * @since 10.4.0
+ * @since 10.5.0
  */
 class Container extends Component implements Contracts\ComponentV2
 {
--- a/src/Discord/Builders/Components/Container.php
+++ b/src/Discord/Builders/Components/Container.php
@@ -19,7 +19,7 @@ namespace Discord\Builders\Components;
  *
  * @since 10.4.0
  */
-class Container extends Component
+class Container extends Component implements Contracts\ComponentV2
 {
     /**
      * Array of components.
--- a/src/Discord/Builders/Components/Container.php
+++ b/src/Discord/Builders/Components/Container.php
@@ -11,6 +11,13 @@
 
 namespace Discord\Builders\Components;
 
+use Discord\Builders\Components\ActionRow;
+use Discord\Builders\Components\Section;
+use Discord\Builders\Components\TextDisplay;
+use Discord\Builders\Components\MediaGallery;
+use Discord\Builders\Components\File;
+use Discord\Builders\Components\Separator;
+
 /**
  * Containers are a new way to group components together.
  * You can also specify an accent color (similar to embeds) and spoiler it.
@@ -55,7 +62,7 @@ class Container extends Component implements Contracts\ComponentV2
     /**
      * Adds a component to the container.
      *
-     * @param Component $component Component to add.
+     * @param ActionRow|Section|TextDisplay|MediaGallery|File|Separator $component Component to add.
      *
      * @throws \InvalidArgumentException Component is not a valid type.
      * @throws \OverflowException        Container exceeds 10 components.
@@ -66,6 +73,10 @@ class Container extends Component implements Contracts\ComponentV2
     {
         if (count($this->components) >= 10) {
             throw new \OverflowException('You can only have 10 components per container.');
+        }
+
+        if (! ( $component instanceof ActionRow || $component instanceof Section || $component instanceof TextDisplay || $component instanceof MediaGallery || $component instanceof File || $component instanceof Separator )) {
+            throw new \InvalidArgumentException('Invalid component type.');
         }
 
         $this->components[] = $component;
--- a/src/Discord/Builders/Components/Container.php
+++ b/src/Discord/Builders/Components/Container.php
@@ -64,19 +64,6 @@ class Container extends Component
      */
     public function addComponent(Component $component): self
     {
-        $allowed = [
-            ActionRow::class,
-            TextDisplay::class,
-            Section::class,
-            MediaGallery::class,
-            Separator::class,
-            File::class,
-        ];
-
-        if (! in_array($component::class, $allowed)) {
-            throw new \InvalidArgumentException('Container can only contain components of type: Action Row, Text Display, Section, Media Gallery, Separator, or File.');
-        }
-
         if (count($this->components) >= 10) {
             throw new \OverflowException('You can only have 10 components per container.');
         }
--- a/src/Discord/Builders/Components/Container.php
+++ b/src/Discord/Builders/Components/Container.php
@@ -1,0 +1,199 @@
+<?php
+
+/*
+ * This file is a part of the DiscordPHP project.
+ *
+ * Copyright (c) 2015-present David Cole <david.cole1340@gmail.com>
+ *
+ * This file is subject to the MIT license that is bundled
+ * with this source code in the LICENSE.md file.
+ */
+
+namespace Discord\Builders\Components;
+
+/**
+ * Containers are a new way to group components together.
+ * You can also specify an accent color (similar to embeds) and spoiler it.
+ *
+ * @link https://discord.com/developers/docs/interactions/message-components#container
+ *
+ * @since 10.4.0
+ */
+class Container extends Component
+{
+    /**
+     * Array of components.
+     *
+     * @var Component[]
+     */
+    private $components = [];
+
+    /**
+     * Accent color for the container.
+     *
+     * @var int|null
+     */
+    private $accent_color;
+
+    /**
+     * Whether the container is a spoiler.
+     *
+     * @var bool
+     */
+    private $spoiler = false;
+
+    /**
+     * Creates a new container.
+     *
+     * @return self
+     */
+    public static function new(): self
+    {
+        return new self();
+    }
+
+    /**
+     * Adds a component to the container.
+     *
+     * @param Component $component Component to add.
+     *
+     * @throws \InvalidArgumentException Component is not a valid type.
+     * @throws \OverflowException        Container exceeds 10 components.
+     *
+     * @return $this
+     */
+    public function addComponent(Component $component): self
+    {
+        $allowed = [
+            Component::TYPE_ACTION_ROW,
+            Component::TYPE_TEXT_DISPLAY,
+            Component::TYPE_SECTION,
+            Component::TYPE_MEDIA_GALLERY,
+            Component::TYPE_SEPARATOR,
+            Component::TYPE_FILE,
+        ];
+
+        if (! in_array($component->getType(), $allowed)) {
+            throw new \InvalidArgumentException('Container can only contain components of type: Action Row, Text Display, Section, Media Gallery, Separator, or File.');
+        }
+
+        if (count($this->components) >= 10) {
+            throw new \OverflowException('You can only have 10 components per container.');
+        }
+
+        $this->components[] = $component;
+
+        return $this;
+    }
+
+    /**
+     * Add a group of components to the container.
+     *
+     * @param Component[] $components Components to add.
+     *
+     * @throws \InvalidArgumentException Component is not a valid type.
+     *
+     * @return $this
+     */
+    public function addComponents(array $components): self
+    {
+        foreach ($components as $component) {
+            $this->addComponent($component);
+        }
+
+        return $this;
+    }
+
+    /**
+     * Sets the components for the container.
+     *
+     * @param Component[] $components Components to set.
+     *
+     * @return $this
+     */
+    public function setComponents(array $components): self
+    {
+        $this->components = $components;
+
+        return $this;
+    }
+
+    /**
+     * Sets the accent color for the container.
+     *
+     * @param int|null $color Color code for the container.
+     *
+     * @return $this
+     */
+    public function setAccentColor(?int $color): self
+    {
+        $this->accent_color = $color;
+
+        return $this;
+    }
+
+    /**
+     * Sets whether the container is a spoiler.
+     *
+     * @param bool $spoiler Whether the container is a spoiler.
+     *
+     * @return $this
+     */
+    public function setSpoiler(bool $spoiler = true): self
+    {
+        $this->spoiler = $spoiler;
+
+        return $this;
+    }
+
+    /**
+     * Returns all the components in the container.
+     *
+     * @return Component[]
+     */
+    public function getComponents(): array
+    {
+        return $this->components;
+    }
+
+    /**
+     * Returns the accent color for the container.
+     *
+     * @return int|null
+     */
+    public function getAccentColor(): ?int
+    {
+        return $this->accent_color;
+    }
+
+    /**
+     * Returns whether the container is a spoiler.
+     *
+     * @return bool
+     */
+    public function isSpoiler(): bool
+    {
+        return $this->spoiler;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function jsonSerialize(): array
+    {
+        $data = [
+            'type' => Component::TYPE_CONTAINER,
+            'components' => $this->components,
+        ];
+
+        if (isset($this->accent_color)) {
+            $data['accent_color'] = $this->accent_color;
+        }
+
+        if ($this->spoiler) {
+            $data['spoiler'] = true;
+        }
+
+        return $data;
+    }
+}
--- a/src/Discord/Builders/Components/Container.php
+++ b/src/Discord/Builders/Components/Container.php
@@ -65,15 +65,15 @@ class Container extends Component
     public function addComponent(Component $component): self
     {
         $allowed = [
-            Component::TYPE_ACTION_ROW,
-            Component::TYPE_TEXT_DISPLAY,
-            Component::TYPE_SECTION,
-            Component::TYPE_MEDIA_GALLERY,
-            Component::TYPE_SEPARATOR,
-            Component::TYPE_FILE,
+            ActionRow::class,
+            TextDisplay::class,
+            Section::class,
+            MediaGallery::class,
+            Separator::class,
+            File::class,
         ];
 
-        if (! in_array($component->getType(), $allowed)) {
+        if (! in_array($component::class, $allowed)) {
             throw new \InvalidArgumentException('Container can only contain components of type: Action Row, Text Display, Section, Media Gallery, Separator, or File.');
         }
 
--- a/src/Discord/Builders/Components/MediaGalleryItem.php
+++ b/src/Discord/Builders/Components/MediaGalleryItem.php
@@ -125,7 +125,7 @@ class MediaGalleryItem implements JsonSerializable
      */
     public function setDescription(?string $description): self
     {
-        if ($description !== null && strlen($description) > 1024) {
+        if ($description !== null && poly_strlen($description) > 1024) {
             throw new \LengthException('Description cannot exceed 1024 characters.');
         }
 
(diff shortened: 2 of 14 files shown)


src/Discord/Builders/MessageBuilder.php
```diff
@@ -373,19 +373,32 @@ public function setReplyTo(?Message $message = null): self
...
     public function addComponent(Component $component): self
     {
-        if (! ($component instanceof ActionRow || $component instanceof SelectMenu)) {
-            throw new \InvalidArgumentException('You can only add action rows and select menus as components to messages. Put your other components inside an action row.');
-        }
-
-        if (isset($this->components) && count($this->components) >= 5) {
-            throw new \OverflowException('You can only add 5 components to a message');
+        // If it's a v2 component, set the v2 flag.
+        if (
```
valzargaming, line 384:
```php
$v2 = [
    Component::TYPE_SECTION,  
    Component::TYPE_TEXT_DISPLAY,
    Component::TYPE_THUMBNAIL,
    Component::TYPE_MEDIA_GALLERY,
    Component::TYPE_SEPARATOR,
    Component::TYPE_FILE,
    Component::TYPE_CONTAINER
];

if (in_array($component->getType(), $v2)) {
    $this->flags = ($this->flags ?? 0) | self::FLAG_V2_COMPONENTS;

    if (isset($this->components) && count($this->components) >= 10) {
        throw new \OverflowException('You can only add 10 components to a v2 message');
    }
} elseif (! ($component instanceof ActionRow || $component instanceof SelectMenu)) {
    throw new \InvalidArgumentException('You can only add action rows and select menus as components to v1 messages. Put your other components inside an action row.');
} elseif (isset($this->components) && count($this->components) >= 5) {
    throw new \OverflowException('You can only add 5 components to a v1 message');
}
```

src/Discord/Builders/Components/UnfurledMediaItem.php
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    /**
+     * Creates a new unfurled media item from an attachment.
+     *
+     * @param string $filename Name of the attachment file.
+     *
+     * @return self
+     */
+    public static function fromAttachment(string $filename): self
```
valzargaming, line 98:
```php
public static function fromAttachment(Attachment|string $filename): self
{
    return $filename instanceof Attachment
        ? self::new('attachment://'.$filename->filename)
        : self::new("attachment://{$filename}");
}
```

Lulalaby:
P neat implementation ❤️ Good job!

valzargaming:
The v2 contract was an excellent idea, and something I'd like to see in other parts of the library. It would certainly cut down on a lost of instances of checking against hardcoded types like this.

src/Discord/Builders/MessageBuilder.php
```diff
@@ -632,6 +638,16 @@ public function setFlags(int $flags): self
...
     }
 
+    /**
+     * Get the current flags of the message.
+     *
+     * @return int
+     */
+    public function getFlags(): int
+    {
+        return $this->flags;
```
SQKo, line 648: the property might be not set yet

```suggestion
     * @return ?int
     */
    public function getFlags(): ?int
    {
        return $this->flags ?? null;
```

src/Discord/Parts/Channel/Message.php
```diff
@@ -181,6 +181,7 @@ class Message extends Part
     public const FLAG_LOADING = (1 << 7);
     public const FLAG_FAILED_TO_MENTION_SOME_ROLES_IN_THREAD = (1 << 8);
     public const FLAG_SUPPRESS_NOTIFICATIONS = (1 << 12);
+    public const FLAG_V2_COMPONENTS = (1 << 15);
```
SQKo, line 184: The official name is `IS_COMPONENTS_V2`

```suggestion
    public const FLAG_IS_V2_COMPONENTS = (1 << 15);
```

please also adjust the referencing code

src/Discord/Builders/MessageBuilder.php
```diff
@@ -373,19 +373,25 @@ public function setReplyTo(?Message $message = null): self
...
+            if (! ($this->flags & Message::FLAG_V2_COMPONENTS)) {
+                $this->flags |= Message::FLAG_V2_COMPONENTS;
+            }
 
-        if (isset($this->components) && count($this->components) >= 5) {
-            throw new \OverflowException('You can only add 5 components to a message');
+            if (isset($this->components) && count($this->components) >= 10) {
+                throw new \OverflowException('You can only add 10 components to a v2 message');
+            }
+        } elseif (! ($component instanceof Components\ActionRow || $component instanceof Components\SelectMenu)) {
```
SQKo, line 391: Please import directly, so it is easy to update/refactor in future (or alias as seen in functions.php)


```
use Discord\Builders\Components\ActionRow;

use Discord\Builders\Components\SelectMenu;
```

```suggestion
        } elseif (! ($component instanceof ActionRow || $component instanceof SelectMenu)) {
```

SQKo (changes requested):

src/Discord/Builders/Components/Thumbnail.php
```diff
@@ -0,0 +1,157 @@
...
+    /**
+     * Sets the media item for the thumbnail.
+     *
+     * @param string $url URL of the media item.
+     *
+     * @return $this
+     */
+    public function setMedia(string $url): self
+    {
+        $this->media = ['url' => $url];
```
SQKo, line 68: The spec actually expects the media to contain UnfurledMediaItem, even though only the `url` for now that will be taken, that is what you can deal with inside jsonSerialize()

Needs to change on the getter too


For example

```suggestion
     * @param UnfurledMediaItem|string $url URL of the media item.
     *
     * @return $this
     */
    public function setMedia(UnfurledMediaItem|string $url): self
    {
        if (is_string($url)) {
            $url = UnfurledMediaItem::new($url);
        }

        $this->media = $url;
```

with this we can anticipate more changes in future

src/Discord/Builders/Components/Separator.php
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * Set the spacing size to large.
+     *
+     * @return $this
+     */
+    public function setSpacingLarge(): self
+    {
+        return $this->setSpacing(self::SPACING_LARGE);
+    }
```
SQKo, line 106:
```suggestion
```

Don't think this is necessary when there is a `setSpacing()` already, we don't want the code to get huge when there is another type of spacing in future

src/Discord/Builders/Components/Section.php
```diff
@@ -0,0 +1,126 @@
...
+     *
+     * @throws \InvalidArgumentException Component is not a TextDisplay.
+     * @throws \OverflowException Section exceeds 3 text components.
+     *
+     * @return $this
+     */
+    public function addComponent(Component $component): self
+    {
+        if (! ($component instanceof TextDisplay)) {
+            throw new \InvalidArgumentException('Section can only contain TextDisplay components. Use setAccessory() for Thumbnail or Button.');
```
SQKo, line 61:
```suggestion
            throw new \InvalidArgumentException('Section can only contain TextDisplay components.');
```

You are not checking wether the `$component` is a `Thumbnail` or `Button` here so it's better not to explicitly say that.

SQKo: On a second though, the spec is telling us to not hardcode this, so we should not validate this yet
```suggestion
            trigger_error('Section may only contain TextDisplay components.', E_USER_NOTICE);
```

src/Discord/Builders/Components/Section.php
```diff
@@ -0,0 +1,126 @@
...
+     *
+     * @param Thumbnail|Button $component Thumbnail or Button component.
+     *
+     * @throws \InvalidArgumentException Component is not a Thumbnail or Button.
+     *
+     * @return $this
+     */
+    public function setAccessory(Thumbnail|Button $component): self
+    {
+        $this->accessory = $component;
```
SQKo, line 85: The spec is telling us to not hard code the supported components yet (anyway leave the phpdoc as is)

```suggestion
    public function setAccessory(Component $component): self
    {
        if (! ($component instanceof Thumbnail || $component instanceof Button)) {
            trigger_error('Accessory may only contain Thumbnail or Button component.', E_USER_NOTICE);
        }
        $this->accessory = $component;
```

src/Discord/Builders/Components/MediaGalleryItem.php
```diff
@@ -0,0 +1,190 @@
...
+     *
+     * @param string|null $description Description for the media item (max 1024 characters).
+     *
+     * @throws \LengthException Description exceeds 1024 characters.
+     *
+     * @return $this
+     */
+    public function setDescription(?string $description): self
+    {
+        if ($description !== null && strlen($description) > 1024) {
```
SQKo, line 128:
```suggestion
        if ($description !== null && mb_strlen($description) > 1024) {
```

multibyte

src/Discord/Builders/Components/Container.php
```diff
@@ -0,0 +1,186 @@
...
+     * @throws \OverflowException        Container exceeds 10 components.
+     *
+     * @return $this
+     */
+    public function addComponent(Component $component): self
+    {
+        if (count($this->components) >= 10) {
+            throw new \OverflowException('You can only have 10 components per container.');
+        }
+
```
SQKo, line 70: might want to validate the component type that can be added here or perhaps strict type them

SQKo (changes requested):
Please also adjust the class implements `Components\xx` to have the namespace imported per my previous suggestion

SQKo:
According to the spec, using the components v2 may disable content and embeds, it's nice to have to warn the users in the MessageBuilder to avoid confusion

src/Discord/Builders/MessageBuilder.php
Exanlv: If it's not set yet you'll also need to coalesce it to null to prevent a warning
```php
return $this->flags ?? null;
```

src/Discord/Builders/MessageBuilder.php
```diff
@@ -373,19 +373,25 @@ public function setReplyTo(?Message $message = null): self
...
+     * @throws \OverflowException        Builder exceeds component limits.
      *
      * @return $this
      */
     public function addComponent(Component $component): self
     {
-        if (! ($component instanceof ActionRow || $component instanceof SelectMenu)) {
-            throw new \InvalidArgumentException('You can only add action rows and select menus as components to messages. Put your other components inside an action row.');
-        }
+        if ($component instanceof ComponentV2) {
```
Exanlv, line 383: I dont think this implementation is correct.

Components v2 can have up to 10 components.
Components v1 can have up to 5.

Every v1 component is also valid for v2 (unless I'm missing something in the docs?).

From what I can tell, with the current implementation, if I have a message with 5 v2 components and I add a v1 component, it'll throw an overflow exception on line 394, regardless of whether the message has the v2 components flag.

`elseif` generally is also more annoying to read IMO, getting rid of them would be nicer.

Something along these lines would be my suggestion (though without the comments, dont copy paste those ofc)
```php
$this->components ??= []; // Removes the need for isset() checks later

// I don't like the auto-flagging for v2 components, I think a `->useV2Components()` method on the message MessageBuilder is better, but if you must, this would be the place
// .. auto adding flag

if (count($this->components) >= 10) { // No need to explicitly check for v2 here, as the count can never go over 5 without being v2 already
    throw new \OverflowException('You can only add 10 components to a v2 message');
}

if (!($this->flags & Message::FLAG_V2_COMPONENTS)) {
    if (count($this->components) >= 5) {
        throw new \OverflowException('You can only add 5 components to a v1 message');
    }

    if (!($component instanceof ActionRow || $component instanceof SelectMenu)) {
        throw new \InvalidArgumentException('You can only add action rows and select menus as components to v1 messages. Put your other components inside an action row.');
    }
}

$this->components[] = $component;
```

To elaborate on my dislike for automagically adding the flag, IMO it should then also handle the following cases:
- 6 v1 components -> convert to v2 components flag?
- A non action row or select menu -> also convert to v2 components flag?

These should be case-by-case, decided by lib user, rather than lib enforcing something. If you can't handle all cases, I don't think you should handle some, gives the user false hope.

---

To what end do we want to validate the flags are valid?

With no additional changes to the `setFlags` method, user could break validation with something along these lines:
```php
$messageBuilder->addComponent($myV2Component)
    ->setFlags(1 << 4);
```
`setFlags` overwrites the automatically added flag, thus Discord will treat it as v1 and return an error.

src/Discord/Builders/Components/UnfurledMediaItem.php
```diff
@@ -0,0 +1,222 @@
...
+    {
+        $data = ['url' => $this->url];
+
+        if ($this->isResolved()) {
+            $resolved = array_filter([
+                'proxy_url' => $this->proxy_url ?? null,
+                'width' => $this->width ?? null,
+                'height' => $this->height ?? null,
+                'content_type' => $this->content_type ?? null,
+                'loading_state' => $this->loading_state ?? null,
```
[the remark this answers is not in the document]
Exanlv: If it's never null, it doesn't need to go through the `array_filter` either and can just be added to the `array_merge`

src/Discord/Builders/Components/UnfurledMediaItem.php
```diff
@@ -0,0 +1,222 @@
...
+            ], fn ($value) => $value !== null);
+
+            return array_merge($data, $resolved);
```
Exanlv, line 217:
```suggestion
            return array_merge($data, $resolved, ['loading_state' => $this->loading_state]);
```

src/Discord/Builders/MessageBuilder.php
valzargaming: Why would adding more than 5 v1 components convert it to a v2 flag? They are neither v2 components nor is that a valid condition for updating the flag. I don't think it's possible to mix v1 and v2 components, or if it is I doubt that it will be on release.

Exanlv: > I don't think it's possible to mix v1 and v2 components, or if it is I doubt that it will be on release.

According to docs it seems like any v1 component will also be a valid v2 component
image

src/Discord/Builders/Components/Separator.php
valzargaming: This certainly makes sense in the context of another library like Laracord, but like you say it should be an external implementation, not the library itself/directly. Ensuring the setSpacing function only accepts valid inputs should be the only responsibility of the library for now.

src/Discord/Builders/MessageBuilder.php
valzargaming: I'm not actually sure this function should exist at all. Flags is a private and otherwise inaccessible property in all other similar parts of the library (member, channel, thread, message, messagebuilder). It's only been expected for the lib user to set the flag, not be checking what it's set to after the fact.

src/Discord/Builders/MessageBuilder.php
valzargaming: Upon further review of the code my opinion has shifted for different reasons. As stated in another comment I made, the flags property is a private variable that does not have a get function in any other implementation in the library. While I believe it should be possible to set the flag, I don't know how it should be properly done within the scope of this class. I definitely think it makes sense for the library to handle it, but the current design pattern calls for either updating the existing MessageBuilder class to accommodate the changes rather than attempting to update the flag within the component class itself.

src/Discord/Builders/MessageBuilder.php
valzargaming: Looking further (again) I think this is an appropriate change to the Builder classes.

src/Discord/Builders/MessageBuilder.php
valzargaming: If any component can be v2, is there a reason we couldn't always just be flagging the class as using v2 and raise the overall component limit to 10? Does the limit automatically increase to 10 as long as there is at least one v2 specific component added? I think these are the questions we need to answer to determine the next steps.

src/Discord/Builders/Components/Container.php
valzargaming: Both PHPDoc and instanceof methods have been implemented.

src/Discord/Builders/MessageBuilder.php
```diff
@@ -632,6 +640,16 @@ public function setFlags(int $flags): self
...
     }
 
+    /**
+     * Get the current flags of the message.
+     *
+     * @return ?int
+     */
+    public function getFlags(): ?int
+    {
+        return $this->flags ?? null;
```
valzargaming, line 650: Would it make more sense to return 0 instead of null?

Log1x: this sounds good

src/Discord/Builders/MessageBuilder.php
Log1x: i think it really just depends if the message contains normal content or an embed, then it can't be v2.

src/Discord/Builders/MessageBuilder.php
valzargaming: Done and updated relevant parts

src/Discord/Builders/Components/UnfurledMediaItem.php
```diff
@@ -0,0 +1,222 @@
...
+     * Sets the URL of the media item.
+     *
+     * @param string $url URL or attachment reference of the media item.
+     *
+     * @return $this
+     */
+    public function setUrl(string $url): self
+    {
+        $this->url = $url;
+
```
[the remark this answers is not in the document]
valzargaming: Do we know if it's the same spec? We could just copy-paste the function here and do the same thing.

src/Discord/Builders/MessageBuilder.php
valzargaming: I reached out to a developer of another lib and confirmed that the v2 limits take place whenever the flag is changed and confirmed the conditions. Setting the v2 flag:
1. Raises the component limit from 5 to 10
2. Disallows the content field
3. Disallows the embeds field

The simplest implementation for this:
1. Make the lib user specify whether or not the message will utilize components v2 by setting the flag
2. Enforce the 5 components by default, allowing up to 10 if the v2 flag is set
3. Update `jsonSerialize()` to drop the content and embeds fields if the v2 flag is set

valzargaming: Updated `jsonSerialize()` as defined above. It's fine for the lib to automatically add the v2 flag if a v2 component is added, and the lib's users can manually set the flag if they'd like to use up to ten v1 components if they're fine with not sending content and embeds.

valzargaming: Updated addComponent to follow the logic flow I described above. I left the ActionRow and SelectMenu exceptions intact for now unless we determine that we should just assume and set the v2 flag otherwise.

src/Discord/Builders/Components/UnfurledMediaItem.php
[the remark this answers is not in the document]
valzargaming: I've confirmed it's exactly the same **_except_** for File, in which case only attachment is allowed.

src/Discord/Builders/Components/Section.php
SQKo: Then don't throw the errors, remove the argument checking, and leave the type `Component` in the parameter while leaving the phpdoc `TextDisplay` for now

src/Discord/Builders/Components/Section.php
SQKo: No, throwing error is the same as hardcoding it, the spec tells us NOT to

if you insist users should be informed then go back with my previous suggestions of `trigger_error()` as the `E_USER_NOTICE` doest not throw error but showing notice tracing to this code in their logs.

valzargaming: The rest of the library does not do this. It should be a separate PR.

valzargaming (approved):

valzargaming (approved):

PR #1295, opened 2025-02-23
Update trigger_error() PHP 8.4 deprecation

SQKo:
now it throws a RuntimeException instead


Exanlv (approved):

Exanlv:
If you don't want people catching it, could also change it to something along these lines:
```php
echo 'My error', PHP_EOL;
exit(1);
```
Can't use catch, and gives you a non-zero exit code.

valzargaming (approved):

valzargaming (approved):

PR #1305, opened 2025-03-21
ExCollectionInterface

valzargaming:
This includes the newly introduced functions that were added as part of recent enhancements. We do not use these functions internally within the library, so rather than update the original interface to include them as requirements, it makes sense for end-users to utilize the ExCollectionInterface instead to take advantage of the full functionality.

For example:
```php
/**
 * Returns the highest role from a collection of roles.
 *
 * @param  Collection<Role> $roles The collection of roles.
 * 
 * @return ?Role
 */
function getHighestRole(Collection $roles): ?Role
{
    return $roles->reduce(fn($prev, $role) =>
        ($prev === null
            ? $role
            : ($this->comparePositionTo($role, $prev) > 0
                ? $role
                : $prev)))
        ->shift(); // Undefined method 'shift', because reduce returns CollectionInterface
}
```

### DIFF
--- a/src/Discord/Helpers/Collection.php
+++ b/src/Discord/Helpers/Collection.php
@@ -17,7 +17,7 @@ namespace Discord\Helpers;
  * @since 5.0.0 No longer extends Laravel's BaseCollection
  * @since 4.0.0
  */
-class Collection implements CollectionInterface
+class Collection implements ExCollectionInterface
 {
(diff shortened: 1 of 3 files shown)


valzargaming:
I've tested and confirmed it fixes the issue, so I'm merging this and pushing a patch release.

PR #1320, opened 2025-04-15
Enhance Attribute and Property Handling for Parts

valzargaming:
This is merely meant as a proof of concept at this time.

#### Summary
This PR introduces enhancements to the `PartTrait` class by adding new methods to improve attribute and property handling. These changes aim to provide better flexibility and maintainability when working with part attributes and properties.

#### Changes Introduced
1. **New Methods Added:**
   - `checkForGetMethod`: Checks if a `get` method exists for a given key.
   - `checkForSetMethod`: Checks if a `set` method exists for a given key.
   - `getProperty`: Retrieves a property using a `get` method if it exists.
   - `setAttributeOrProperty`: Sets an attribute or a property, supporting mutators and methods.

2. **Enhancements to Magic Methods:**
   - Updated `__get` to use `getProperty` as a fallback if the attribute does not exist.
   - Updated `__set` to utilize `setAttributeOrProperty` for setting values.

3. **Improved Code Maintainability:**
   - Consolidated logic for handling attributes and properties in a consistent and reusable manner.

#### Example Usage
```php
$part->discord; // Will attempt to use getDiscord() if it exists
$part->color = (0xE1452D); // Will attempt to use setColor() if it exists
```

### DIFF
--- a/src/Discord/Parts/PartTrait.php
+++ b/src/Discord/Parts/PartTrait.php
@@ -114,6 +114,24 @@ trait PartTrait
     }
 
     /**
+     * Checks if a 'get' method exists for the given key.
+     *
+     * @param string $key The property name to check.
+     *
+     * @return string|false Either a string if it is a method or false.
+     */
+    private function checkForGetMethod(string $key): string|false
+    {
+        $str = 'get'.self::studly($key);
+
+        if (method_exists($this, $str)) {
+            return $str;
+        }
+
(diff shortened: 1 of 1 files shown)


valzargaming:
I should probably mention that these are a backwards compatible way using magic methods. Property hooks are the way forward, but we can't utilize them in the current minimum PHP version.

PR #1329, opened 2025-05-19
Automatically wrap SelectMenus in ActionRow

valzargaming:
The old behavior did this directly in [SelectMenu::jsonSerialize](https://github.com/discord-php/DiscordPHP/blob/6ed0ae1a62b105ec1553cc3232ebaae3144e7d6b/src/Discord/Builders/Components/SelectMenu.php#L530), such that anyone adding a SelectMenu manually to an ActionRow (as is the correct usage) would end up with an ActionRow nested inside of another ActionRow, which would always result in Discord rejecting with a `50035 Invalid Form Body` error message.

The new behavior moves this logic directly to Container::addComponent and MessageBuilder::addComponent by creating a new ActionRow and adding the SelectMenu to it, preserving the original usage without breaking containers.

### DIFF
--- a/src/Discord/Builders/Components/ActionRow.php
+++ b/src/Discord/Builders/Components/ActionRow.php
@@ -57,6 +57,14 @@ class ActionRow extends Component
             throw new \InvalidArgumentException('You cannot add another `ActionRow` to this action row.');
         }
 
+        if ($component instanceof SelectMenu) {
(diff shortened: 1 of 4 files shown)


valzargaming:
Moved to https://github.com/discord-php/DiscordPHP/pull/1334

PR #1334, opened 2025-05-19
Component Style Classes

valzargaming:
https://discord.com/developers/docs/components/reference#component-object

Objects found in the list with types should have their own extendable `ComponentObject` class, which extends the `Component` class, and indicates that they are guaranteed to have an associated type and may optionally contain an id. Additionally, the new `Interactive` object extends `ComponentObject`s and adds the optional `custom_id` field.

New objects `Layout` and `Content` have also been created, though they serve no functional purpose at this time other than to distinguish them from other types of `ComponentObject`s as outlined in the official documentation.

The library currently treats some components objects (e.g. `UnfurledMediaItem`) as though they are a traditional component, so to prevent accidental errors when attempting to call the new `ComponentObject::getType` or `ComponentObject::getId` methods, default method have been implemented on the base `Component` class and are overwritten within the new `ComponentObject` class. This may change with the final 10.9.0 release.


valzargaming:
The base Component class is being marked as @deprecated but may still be used internally within the library, though those instances will soon be cut down to utilize the newer classes.

valzargaming:
The ComponentObject table contains the appropriate usage, either Message or Modal, and that has now been defined as a const with an array. If you have an application where you may be dynamically building responses from components (such as from json content), you can now refer to these const values to determine if the component you're attempting to use is appropriate for use in your response.

valzargaming:
Git be hard sometimes

valzargaming:
Tested by @jocafamaka and confirmed working, thanks for the assist!

PR #1341, opened 2025-05-22
Collectibles, Nameplate, and Primary Guild

valzargaming:
https://github.com/discord/discord-api-docs/pull/7506/files


valzargaming:
Tested and confirmed working, however `primaryGuild` doesn't seem to be visible to bot users yet. Will update in future patch if things change, but for now I've used json data that I can see as a user to confirm that it should work.

PR #1345, opened 2025-05-31
Component part

### LINKED ISSUE #1344: Message components cannot be retrieved from Interaction context
**Environment**
- PHP Version:
    - 8.4.7
- DiscordPHP Version:
    - 10.10.12

**Describe the bug**
Message components cannot be accurately retrieved from an interaction.

**To Reproduce**
```php
`var_dump($interaction->message);`
```
You'll notice that components is a Collection containing Discord\Parts\Interactions\Request\Component objects, but the only populated field is type.

**Expected behavior**
Component should be retrievable with all of their data intact.

valzargaming:
Closes #1344

### DIFF
--- a/src/Discord/Parts/Channel/Message.php
+++ b/src/Discord/Parts/Channel/Message.php
@@ -720,7 +720,7 @@ class Message extends Part
     /**
      * Returns the components attribute.
      *
-     * @return ExCollectionInterface|Component[]|null $components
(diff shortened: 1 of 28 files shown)


valzargaming:
All parts have been tested and confirmed working except for InteractionData's as I don't have any interactions that uses modals to with. It uses the same code as is used within the rest of the library, so I don't see why it wouldn't work.

PR #1350, opened 2025-06-04
Static methods for making link and premium buttons

valzargaming:


valzargaming:
Also fixed the outdated checking of styles in __construct

PR #1359, opened 2025-06-13
Remove ext-json requirement

norkunas:
As of PHP 8.0, json is bundled into core


norkunas:
I also need to ask - why bloated carbon library is required here? I see it's only to parse data from discord to create carbon objects, but that's too much imho.. simple DateTimeImmutable would be enough for many cases, and for advanced users could require it themself..

valzargaming:
Tagging Brandon in this as I'm uncertain if this will affect downstream frameworks

Log1x (approved):

Log1x:
LGTM

PR #1367, opened 2025-07-05
ChannelTrait

valzargaming:


valzargaming:
I've been testing this branch with my bot and haven't run into any issues. I'm merging this with since tag set to 10.19.0 instead of the new patch of 10.18.5 to give time for any issues to make themselves known before making it officially supported.

PR #1371, opened 2025-07-19
Clickable Links and Customizable Statuses in Rich Presence

valzargaming:
https://discord.com/developers/docs/change-log#clickable-links-and-customizable-statuses-in-rich-presence


valzargaming:
Never mind about the testing. Apparently the data that bots receive on the presence update event is very limited and not the same as what users see. Going to go ahead and merge this, please shout at me if you find something broken.

PR #1381, opened 2025-08-08
Refactor chunking logic

valzargaming:
This pull request refactors the chunking logic in the `Discord.php` file to improve code organization and maintainability. The main changes involve extracting anonymous functions into named protected methods, which clarifies the responsibilities and flow of the chunking process.

**Refactoring and code organization:**

* Extracted the anonymous `$checkForChunks` function into a new protected method `checkForChunks`, and updated the periodic timer to call this method directly.
* Extracted the anonymous `$sendChunks` function into a new protected method `sendChunks`, and updated the logic to use this method for sending guild member chunks.
* Replaced inline function calls with direct calls to the new protected methods, simplifying the setup and invocation of chunking and chunk sending.

### DIFF
--- a/src/Discord/Discord.php
+++ b/src/Discord/Discord.php
@@ -1033,60 +1033,70 @@ class Discord
             return $this->ready();
         }
 
-        $checkForChunks = function () {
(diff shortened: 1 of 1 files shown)


valzargaming:
Tested and confirmed working.

PR #1398, opened 2025-09-04
Bump Version:)

skyss0fly:


valzargaming:
We only really bump the constant variable when it would help with troubleshooting support issues, otherwise it's just one more thing to remember to update when making updates. We're holding out until 10.19 is ready to reduce the amount of overhead.

That said, .64 is the current patch version, and this PR wouldn't go live until .65, so this would need to be bumped again.

PR #1416, opened 2025-10-06
Hand over the matched option value into the autocomplete callback

codemasher:
This saves digging through the entire `Interaction` instance again in order to find the `focused` value that triggered the callback.

### DIFF
--- a/src/Discord/WebSockets/Events/InteractionCreate.php
+++ b/src/Discord/WebSockets/Events/InteractionCreate.php
@@ -103,7 +103,7 @@ class InteractionCreate extends Event
      *
      * @return bool Returns true if a suggestion was triggered, otherwise false.
      */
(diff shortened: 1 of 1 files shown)


codemasher:
I'm taking out the thoughts from the OP and leave them here, this is material for another day as there are currently too many open questions re this:

> I'd also like to hand over the matched command/subcommand/subcommandgroup, for which I'd like to dissolve that recursive call (recursions are a bit overused in your library anyway for my taste :D). From what I gather, the options instance is nested similar to [the description here](https://discord.com/developers/docs/interactions/application-commands#subcommands-and-subcommand-groups), or is there something I'm overlooking?
3 nested foreach calls might look a bit clunky, but are probably faster and leave the option to hand over the command names too.

re OP: LGTM?

PR #1422, opened 2025-10-15
Reaction class should use Emoji Part rather than stdClass

valzargaming:
This pull request makes a minor improvement to the `Reaction` class in the Discord package by initializing the `emoji` attribute with an `Emoji` object instead of a generic `stdClass`. This change enhances type safety and consistency when handling emoji data.

- Improved emoji initialization:
  * In the `setIdAttribute` method of `Reaction.php`, replaced the initialization of the `emoji` attribute from a `stdClass` to an `Emoji` object for better type safety and clarity.
  * Removed the unused `stdClass` import from `Reaction.php`.

### DIFF
--- a/src/Discord/Parts/Channel/Reaction.php
+++ b/src/Discord/Parts/Channel/Reaction.php
@@ -22,7 +22,6 @@ use Discord\Parts\Part;
 use Discord\Parts\Thread\Thread;
 use Discord\Parts\User\User;
 use React\Promise\PromiseInterface;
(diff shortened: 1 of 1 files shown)


valzargaming:
This PR might be pointless thankss to the getEmojiAttribute's attributePartHelper, but should probably be merged anyway just to satisfy the expectation of an Emoji being the value.

alexandre433 (approved):

PR #1430, opened 2025-11-09
Types: Member::$roles

liamduckett:
Hey, first contribution here - thanks for all the existing work!

From my understanding, this property cannot possible be an array:

```php
/**
 * Returns the roles attribute.
 *
 * @return ExCollectionInterface<?Role> A collection of roles the member is in. null role only contains ID in the collection.
 */
protected function getRolesAttribute(): ExCollectionInterface
```


valzargaming:
I think the best way forward is to open an issue and tag this pr in it somewhere. It's going to be a while before I can dive into this more, but I don't want it to get lost.

PR #1452, opened 2025-12-26
Require discord-php-helpers/collection ^8.0

valzargaming:
This pull request removes the local implementation of the `Collection` class and its related interfaces from the codebase, and instead adds a dependency on the external `discord-php-helpers/collection` package (which we also maintain). This change helps to reduce maintenance overhead and ensures consistency by using a shared, externally maintained collection library.

Dependency update:

* Added `discord-php-helpers/collection` as a required dependency in `composer.json` to replace the local collection implementation.

Code cleanup (removal of local collection code):

* Removed the local `Collection` class from `src/Discord/Helpers/Collection.php`.
* Removed the local `CollectionTrait` from `src/Discord/Helpers/CollectionTrait.php`.
* Removed the local `CollectionInterface` from `src/Discord/Helpers/CollectionInterface.php`.
* Removed the local `ExCollectionInterface` from `src/Discord/Helpers/ExCollectionInterface.php`.


valzargaming:
Tested and confirmed working as intended

PR #1466, opened 2026-04-12
Add 'values' to the $fillable array in StringSelect.php

oliverschloebe:
Add 'values' to $fillable in StringSelect.php so that modal submission values survive hydration.

Without this, $component->values returns null because the fill() method silently drops keys not in $fillable.


valzargaming:
Closing this, will be fixed in #1466

valzargaming:
#1467 *

PR #1468, opened 2026-04-19
Add infection/infection as a testing framework

Refaltor77:
## Summary

This PR fixes two critical logic bugs discovered during an internal audit of the codebase, and upgrades the test tooling used to validate them.

### 🐛 Bug 1 — `isInviter` permission check silently disabled (operator precedence)

`Invite::getTargetUsers`, `Invite::updateTargetUsers`, `Invite::updateTargetUsersFromContent`, `Invite::getTargetUsersJobStatus` and `Channel::sendSoundboardSound` contained expressions like:

```php
! $this->inviter->id === $this->discord->user->id
```

Because `!` has higher precedence than `===`, this parsed as `(!$this->inviter->id) === $this->discord->user->id`, which is always `false === <string>` → always `false`. The "is inviter" exemption could therefore never be satisfied, and the voice-channel-match check in `sendSoundboardSound` always failed.

**Fix:** replace with strict inequality `$this->inviter->id !== $this->discord->user->id`.

### 🐛 Bug 2 — `Member::kick()` returned a raw Exception instead of a rejected Promise

In `Member::kick()` the missing-guild branch was:

```php
if (null === $guild) {
    return new \RuntimeException('Member has no Guild Part');
}
```

The surrounding `Member::ban()` already uses `reject(...)` correctly (line 172). Because `kick()` returned a raw `Throwable` as the resolved value, callers using `->then(...)->catch(...)` never saw the error — the Promise resolved silently with the exception object.

**Fix:** wrap in `reject(new \RuntimeException(...))`.

### 🧰 Chore — test tooling

To validate the fixes and back the refactor surface with static analysis + mutation coverage:

- Migrate `phpunit.xml` from the PHPUnit 9/10 schema to PHPUnit 12 (drop deprecated attributes, switch `<coverage><include>` to `<source><include>`).
- Add `phpstan/phpstan ^2.1` and `infection/infection ^0.32` as dev dependencies, plus a minimal `infection.json5` configuration.
- Gitignore `.papline/` (local workflow log directory).

## Test plan

- [x] New unit tests `tests/Parts/Channel/InviteTest.php` (17 tests covering regression + adversarial cases: numeric coercion, leading zeros, zero-width Unicode, 1 MiB ids, 100 concurrent invocations).
- [x] New unit tests `tests/Parts/User/MemberTest.php` (6 tests: null-guild rejection, permission rejection, empty/Unicode reasons, 100 concurrent calls, explicit guard that `kick()` never resolves with a `Throwable` value).
- [x] `./vendor/bin/phpunit` — 66 tests pass (baseline of 1F + 6E preserved, all preexisting integration-test noise unrelated to this change).
- [x] `./vendor/bin/pint --test` — pass on modified files.
- [x] `./vendor/bin/mago lint` — 0 regression vs baseline on modified files.
- [x] `./vendor/bin/phpstan --level=max` — reduced total error count on `Invite.php`+`Channel.php` from 175 → 166, `Member.php` unchanged at 96 (zero regression introduced).
- [x] `./vendor/bin/infection` — MSI 93% on `Invite.php`/`Channel.php` scope, MSI 100% on `Member.php` scope (minimum threshold 85%).

## Note

Authored from a local workflow that runs Pint + Mago + PHPUnit + PHPStan + Infection with automatic rollback on regression. Happy to split the chore commit out into a separate PR if that is preferred.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ phpunit.log
 /.phpunit*
 /coverage
 /.tools
+.papline/
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,9 @@
         "symfony/cache": "^5.4",
         "laravel/pint": "^1.21",
         "pestphp/pest": "^4.1.2",
-        "carthage-software/mago": "1.1.0"
+        "carthage-software/mago": "1.1.0",
+        "phpstan/phpstan": "^2.1",
+        "infection/infection": "^0.32.6"
     },
     "autoload": {
         "files": [
@@ -82,7 +84,8 @@
     "config": {
         "allow-plugins": {
             "pestphp/pest-plugin": true,
-            "carthage-software/mago": true
+            "carthage-software/mago": true,
+            "infection/extension-installer": true
         }
     }
 }
--- a/infection.json5
+++ b/infection.json5
@@ -1,0 +1,16 @@
+{
+    "$schema": "vendor/infection/infection/resources/schema.json",
+    "source": {
+        "directories": ["src"]
+    },
+    "timeout": 30,
+    "logs": {
+        "text": ".papline/infection.log",
+        "summary": ".papline/infection-summary.log"
+    },
+    "mutators": {
+        "@default": true
+    },
+    "testFramework": "phpunit",
+    "testFrameworkOptions": "--filter=InviteTest"
+}
--- a/infection.json5
+++ b/infection.json5
@@ -12,5 +12,5 @@
         "@default": true
     },
     "testFramework": "phpunit",
-    "testFrameworkOptions": "--filter=InviteTest"
+    "testFrameworkOptions": "--filter=\"(InviteTest|MemberTest)\""
 }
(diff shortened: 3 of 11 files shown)


valzargaming:
Please split bug fixes and enhancements to unit tests into separate PRs

Refaltor77:
Closing this PR — after rebasing onto the latest `master`, the three bug fixes proposed here turned out to be already present upstream:

- `Channel::sendSoundboardSound` precedence fix → merged as `8258c2ff Fix operator`
- `Invite` precedence fix (4 sites) → merged as `565e53ea Fix operators`
- `Member::kick()` null-guild `reject()` wrapping → merged as `35f15b65 RuntimeException => PartRequestFailedException`

Per your review comment ("split bug fixes and enhancements to unit tests into separate PRs"), I'll re-open a fresh PR containing **only** the test tooling upgrade (PHPUnit 12 config + phpstan + infection) and the new unit tests (`InviteTest`, `MemberTest`) that lock in those fixes against regression.

Thanks for the review!

PR #1469, opened 2026-04-20
Upgrade test tooling to PHPUnit 12 + add Infection/PHPStan + regression tests

Refaltor77:
## Summary

Follow-up to #1468 (closed) — that PR mixed bug fixes (now all merged upstream via `8258c2ff`, `565e53ea`, `35f15b65`) with tooling/tests. Per the review comment asking to split bug fixes and test enhancements into separate PRs, this PR contains **only** the test-tooling upgrade and the unit tests that lock in those three fixes.

Two commits, each reviewable independently:

### 1. `chore(test): upgrade phpunit.xml to PHPUnit 12 + add phpstan/infection`

- Migrate `phpunit.xml` from the PHPUnit 9/10 schema to PHPUnit 12 (drop deprecated attributes, switch `<coverage><include>` to `<source><include>`, unified `cacheDirectory`).
- Add `phpstan/phpstan ^2.1` and `infection/infection ^0.32` as dev dependencies + a minimal `infection.json5` config scoped for fast runs.
- Gitignore `.papline/` (local workflow log directory).

### 2. `test: add InviteTest / MemberTest regression + adversarial suites`

- `tests/Parts/Channel/InviteTest.php` — 17 tests covering the `isInviter` precedence regression (bot-is-inviter bypass, non-inviter rejection, `manage_guild` / `view_audit_log` exemptions) plus adversarial cases (leading zeros, trailing space, zero-width space, 1 MiB ids, 100 concurrent invocations).
- `tests/Parts/User/MemberTest.php` — 6 tests covering `Member::kick()` null-guild path (regression: must return a rejected Promise, not resolve with a Throwable value), missing-permission path, empty/Unicode reasons, 100 concurrent calls.

## Test plan

- [x] `./vendor/bin/phpunit` — new tests pass against current `master`
- [x] `./vendor/bin/pint --test` — pass on modified files
- [x] `./vendor/bin/mago lint` — zero regression vs baseline
- [x] `./vendor/bin/phpstan --level=max` — zero regression on touched files
- [x] `./vendor/bin/infection` — MSI 93% on `Invite.php`/`Channel.php` scope, MSI 100% on `Member.php` scope (threshold 85%)

## Note

No source files under `src/` are modified in this PR. All three underlying fixes these tests cover are already present on `master`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ phpunit.log
 /.phpunit*
 /coverage
 /.tools
+.papline/
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,9 @@
         "symfony/cache": "^5.4",
         "laravel/pint": "^1.21",
         "pestphp/pest": "^4.1.2",
-        "carthage-software/mago": "1.1.0"
+        "carthage-software/mago": "1.1.0",
+        "phpstan/phpstan": "^2.1",
+        "infection/infection": "^0.32.6"
     },
     "autoload": {
         "files": [
@@ -82,7 +84,8 @@
     "config": {
         "allow-plugins": {
             "pestphp/pest-plugin": true,
-            "carthage-software/mago": true
+            "carthage-software/mago": true,
+            "infection/extension-installer": true
         }
     }
 }
--- a/infection.json5
+++ b/infection.json5
@@ -1,0 +1,16 @@
+{
+    "$schema": "vendor/infection/infection/resources/schema.json",
+    "source": {
+        "directories": ["src"]
+    },
+    "timeout": 30,
+    "logs": {
+        "text": ".papline/infection.log",
+        "summary": ".papline/infection-summary.log"
+    },
+    "mutators": {
+        "@default": true
+    },
+    "testFramework": "phpunit",
+    "testFrameworkOptions": "--filter=\"(InviteTest|MemberTest)\""
+}
--- a/tests/Parts/Channel/InviteTest.php
+++ b/tests/Parts/Channel/InviteTest.php
@@ -13,11 +13,62 @@ declare(strict_types=1);
  */
 
 use Discord\Discord;
-use Discord\Parts\Channel\Invite;
 use Discord\Http\Endpoint;
+use Discord\Http\Exceptions\NoPermissionsException;
+use Discord\Parts\Channel\Channel;
+use Discord\Parts\Channel\Invite;
+use Discord\Parts\Permissions\RolePermission;
+use Discord\Parts\User\User;
 use PHPUnit\Framework\TestCase;
+use React\Promise\PromiseInterface;
 
 use function React\Promise\resolve;
+
+/**
+ * Test stub for Invite that overrides the protected attribute getters so
+ * tests can inject stubbed Channel and User parts without needing the full
+ * factory / repository chain.
+ */
+class InviteTestStub extends Invite
+{
+    public ?Channel $_channelStub = null;
+    public ?User $_inviterStub = null;
+
+    protected function getChannelAttribute(): ?Channel
+    {
+        return $this->_channelStub;
+    }
+
+    protected function getInviterAttribute(): ?User
+    {
+        return $this->_inviterStub;
+    }
+}
+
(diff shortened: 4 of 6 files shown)


tests/Parts/Channel/InviteTest.php
```diff
@@ -13,12 +13,63 @@
...
+use React\Promise\PromiseInterface;
 
 use function React\Promise\resolve;
 
+/**
+ * Test stub for Invite that overrides the protected attribute getters so
+ * tests can inject stubbed Channel and User parts without needing the full
+ * factory / repository chain.
+ */
+class InviteTestStub extends Invite
```
alexandre433, line 32: These "stubs" should not exist as a class inside the same file that the test is in.

Either create a "fake" class with `new <class>()` or add these classes to separate files; If you choose to create new files, I'd suggest adding into a folder called "Stubs".

tests/Parts/Channel/InviteTest.php
```diff
@@ -53,4 +104,498 @@ public function testUpdateTargetUsersSendsMultipartPut()
...
+     * Before the fix the condition `! $this->inviter->id === $this->discord->user->id`
+     * was parsed as `(! $this->inviter->id) === $this->discord->user->id`, which is
+     * always `false === <string>` → always `false`, so the inviter clause could
+     * never exempt the bot from the perms reject.
+     *
+     * After the fix (`$this->inviter->id !== $this->discord->user->id`), a bot
+     * that is the inviter is allowed through even when it lacks manage_guild
+     * and view_audit_log.
+     */
+    public function testGetTargetUsersAllowsBotWhenBotIsInviter(): void
```
alexandre433, line 121: Why the comment on this test and not the others?

tests/Parts/Channel/InviteTest.php
alexandre433, line 1: Big advice here; from what I see in the current main code, the tests are not properly formed, if you want your tests to properly work, make the base work global, as in ready to get picked up by other classes, but that doesn't overwrite what already exists.

tests/Parts/User/MemberTest.php
```diff
@@ -0,0 +1,310 @@
...
+use PHPUnit\Framework\TestCase;
+use React\Promise\PromiseInterface;
+
+use function React\Promise\resolve;
+
+/**
+ * Test stub that lets us instantiate Member without running the full Part
+ * constructor (which requires a fully-wired Discord instance).
+ */
+class MemberTestStub extends Member
```
alexandre433, line 29: ?

tests/Parts/User/MemberTest.php
alexandre433, line 1: Same as before with the testing structure, stubs and comments.

.gitignore
```diff
@@ -30,3 +30,4 @@ phpunit.log
 /.phpunit*
 /coverage
 /.tools
+.papline/
```
alexandre433, line 33: Follow other files'/folders' structure `/.papline`

composer.json
```diff
@@ -46,7 +46,9 @@
         "symfony/cache": "^5.4",
         "laravel/pint": "^1.21",
         "pestphp/pest": "^4.1.2",
-        "carthage-software/mago": "1.1.0"
+        "carthage-software/mago": "1.1.0",
+        "phpstan/phpstan": "^2.1",
+        "infection/infection": "^0.32.6"
```
alexandre433, line 51: Remove unnecessary packages;
phpstan is already imported via `laravel/pint` which imports `larastan/larastan` which imports `phpstan`.

infection.json5
alexandre433, line 1: Not required; remove.

alexandre433 (changes requested):
It just looks like AI slop; without a care in the current project's structure, psr-12 usage, and adding tools just willy nilly without really checking what we really are using already.

tests/Parts/User/MemberTest.php
valzargaming: This makes even less sense to me because the test suite does have support for creating a mock instance of the Discord client, so it's already possible to do (and other tests do this currently)

alexandre433:
Converting to draft.

PR #1470, opened 2026-04-20
fix: Embed::addField off-by-one (26th field accepted) + centralised payload limits

Refaltor77:
## Summary

- **Bug fix**: \`Embed::addField()\` rejects only when the embed already has *more than* 25 fields (\`count > 25\`), so a 26th field is accepted and Discord replies with HTTP 400. Tightened to \`>= EMBED_FIELDS_MAX\` so the 26th call throws up front.
- Added the per-field validations the docstring already promised: field name ≤ 256 chars, field value ≤ 1024 chars, combined embed text ≤ 6000 chars.
- Introduced \`Discord\\Helpers\\ValidatesDiscordLimits\` — an interface that centralises Discord's documented payload limits (embed + message) as constants, replacing hard-coded literals scattered across \`Embed.php\`. Other builders can adopt it incrementally.

## Test plan

- [x] \`./vendor/bin/pint --test\` — pass on modified files
- [x] \`./vendor/bin/phpstan --level=max\` — zero regression on touched files (59 → 59)
- [x] \`./vendor/bin/mago lint\` — zero regression vs baseline

## Note

Per the split-PRs review feedback, this PR contains **only the source fix**. A follow-up PR will add the regression + adversarial unit tests (16 tests, Infection MSI 86%) against \`Embed.php\` once this lands.

### DIFF
--- a/src/Discord/Helpers/ValidatesDiscordLimits.php
+++ b/src/Discord/Helpers/ValidatesDiscordLimits.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is a part of the DiscordPHP project.
+ *
+ * Copyright (c) 2015-2022 David Cole <david.cole1340@gmail.com>
+ * Copyright (c) 2020-present Valithor Obsidion <valithor@discordphp.org>
+ *
+ * This file is subject to the MIT license that is bundled
+ * with this source code in the LICENSE.md file.
+ */
+
+namespace Discord\Helpers;
+
+/**
+ * Centralises Discord API payload size limits as shared constants.
+ *
+ * Classes that produce or validate Discord-bound payloads implement this
+ * interface to expose consistent limits in a single location. Grouping
+ * them prevents the "magic number" drift that happens when the same
+ * limit is repeated across builders and parts.
+ *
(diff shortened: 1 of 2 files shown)


src/Discord/Helpers/ValidatesDiscordLimits.php
alexandre433, line 1: Wrong type of file (the interface itself); In php, if you don't know, there are other types of "classes" like `Traits` or `Enum`'s - please use the proper one.

src/Discord/Helpers/ValidatesDiscordLimits.php
```diff
@@ -0,0 +1,64 @@
...
+    public const EMBED_TOTAL_CHARS_MAX = 6000;
+
+    /** Maximum characters in a message content field. */
+    public const MESSAGE_CONTENT_MAX = 2000;
+
+    /** Maximum embeds attached to a single message. */
+    public const MESSAGE_EMBEDS_MAX = 10;
+
+    /** Maximum file attachments on a single message. */
+    public const MESSAGE_FILES_MAX = 10;
```
alexandre433, line 63: Some values are duplicated, if you're just going for naming, add both into 1 or add a function that calls the same value.

alexandre433 (changes requested):

alexandre433:
Converting to draft.

PR #1472, opened 2026-04-20
Add AI Skills Infrastructure

splashsky:
Just some handy dandy files that add "skills" for AI agents to read to (hopefully) make AI workflows more consistent across the repo.

At the top level:
 
- `SKILLS.md` is the quick map. It lays out the big-picture rules for the repo and points agents at the right skill depending on what they’re touching.
- `AGENTS.md` is more of a working guide. It explains how the repo is structured, which files tend to move together, and the kinds of boundaries agents should try to preserve.
 
Then there are a few deeper skills for the main hotspots:
 
- `skills/part-model-maintainer.md` covers how Parts are supposed to work: `$fillable`, mutators, nested typed data, repository bindings, save/fetch hooks, and docblocks.
- `skills/repository-cache-keeper.md` covers repositories: endpoint vars, CRUD flow, cache behavior, and where repository logic should stop and part logic should take over.
- `skills/gateway-cache-sync-keeper.md` covers gateway event handlers: how payloads become typed parts, how caches should be updated, and what related repos or side effects need to stay in sync.
- `skills/builder-payload-smith.md` covers builders: payload validation, `jsonSerialize()` rules, component usage, and keeping builders separate from Parts.
- `skills/interaction-flow-keeper.md` covers interactions: typed interaction handling, resolved data hydration, command/autocomplete routing, and response flow.
- `skills/legacy-command-client-keeper.md` covers the older prefix-command layer so agents keep that model separate from slash-command and interaction code.
 
So basically: broad repo rules up top, then more detailed guidance for the areas that are easiest to get subtly wrong.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,313 @@
+# DiscordPHP Agent Guide
+
+This file is the repo operating manual for AI agents. Use it together with `SKILLS.md`.
+
+- `SKILLS.md` tells you **which specialist mindset to load** for a task.
+- `AGENTS.md` tells you **how to work inside this repository without breaking its design**.
+
+If a change crosses layers, load multiple skills and use the playbooks in this file to keep boundaries clean.
+
+## Start here
+
+Before changing anything:
+
+1. Identify the layer you are touching.
+2. Read the base abstraction for that layer before copying a concrete class.
+3. Read one representative concrete implementation from the same family.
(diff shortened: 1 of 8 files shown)


valzargaming:
Even if this does nothing more than get bots to stop creating some of the worst PRs imaginable I'll take it.

